PR #15, opened 2019-11-19
Update README.md

IamDZN:
Clonning command was missing ".git" part


saadpasta:
@allcontributors add @IamDZN for doc

PR #23, opened 2019-11-26
Update fonts to Montserrat, add button transitions

kartikch918:
Add new font Montserrat (Spotify style font)
https://fonts.google.com/specimen/Montserrat


saadpasta:
@kartikch918 Can You Show Some Screenshots

kartikch918:
New font preview:
Screenshot (51)

src/components/button/Button.css
```diff
@@ -14,7 +15,11 @@
   font-size: 16px;
   cursor: pointer;
 }
-
+.main-button:hover {
+  background-color: #ffffff;
+  color: black;
+  transition: ease-in 0.3s;
+}
```
saadpasta: Do you think this hover effect is necessary

kartikch918: In my opinion, yes. Some simple lightweight transitions can make a page feel a lot smoother and interactive. Also it fits in with the theme too but it's your repo so it's your call mate.
Remove the commit if you don't want to add it.

saadpasta: Bro's it's your Portfolio also it's your choice also and i think you made your point clear and I love it 
Thanks For Contributing

PR #42, opened 2020-01-11
Adds new component for resume abstraction

### LINKED ISSUE #40: Add appropriate href in contact
- Add `tel:` and `mailto:` hrefs for contact section.

### LINKED ISSUE #41: Resume URL abstraction
- Create a new _Resume_ component with iframe src as the drive link for the file, so that it looks cleaner instead of redirecting to Google Drive.

rajkumaar23:
- Fixes #41 

Screenshots :
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+.idea/
--- a/package-lock.json
+++ b/package-lock.json
@@ -5269,7 +5269,8 @@
             },
             "ansi-regex": {
               "version": "2.1.1",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "aproba": {
               "version": "1.2.0",
@@ -5287,11 +5288,13 @@
             },
             "balanced-match": {
               "version": "1.0.0",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "brace-expansion": {
               "version": "1.1.11",
               "bundled": true,
+              "optional": true,
               "requires": {
                 "balanced-match": "^1.0.0",
                 "concat-map": "0.0.1"
@@ -5304,15 +5307,18 @@
             },
             "code-point-at": {
               "version": "1.1.0",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "concat-map": {
               "version": "0.0.1",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "console-control-strings": {
               "version": "1.1.0",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "core-util-is": {
               "version": "1.0.2",
@@ -5415,7 +5421,8 @@
             },
             "inherits": {
               "version": "2.0.3",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "ini": {
               "version": "1.3.5",
@@ -5425,6 +5432,7 @@
             "is-fullwidth-code-point": {
               "version": "1.0.0",
               "bundled": true,
+              "optional": true,
               "requires": {
                 "number-is-nan": "^1.0.0"
               }
@@ -5437,17 +5445,20 @@
             "minimatch": {
               "version": "3.0.4",
               "bundled": true,
+              "optional": true,
               "requires": {
                 "brace-expansion": "^1.1.7"
               }
             },
             "minimist": {
               "version": "0.0.8",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "minipass": {
               "version": "2.3.5",
               "bundled": true,
+              "optional": true,
               "requires": {
                 "safe-buffer": "^5.1.2",
                 "yallist": "^3.0.0"
@@ -5464,6 +5475,7 @@
             "mkdirp": {
               "version": "0.5.1",
               "bundled": true,
+              "optional": true,
               "requires": {
                 "minimist": "0.0.8"
               }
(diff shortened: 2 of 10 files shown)


rajkumaar23:
@saadpasta Kindly look into this PR and suggest changes if any :)

rajkumaar23:
@saadpasta Kindly look into this PR and suggest changes if any :)

saadpasta:
@rajkumaar23 please make 2 separate merge request for href and new resume component so we have a better idea

saadpasta:
@kartik918 what do you think about new resume component

rajkumaar23:
> @rajkumaar23 please make 2 separate merge request for href and new resume component so we have a better idea

I have made two separate 2 PRs. Please check the other one at #44

rajkumaar23:
> Please remove .gitignore, package-lock.json and package.json from your pull request. I guess they are not needed for this Merge Request

- I have added `react-router` for routing, because of which I've committed package files.
- And the .gitignore misses a folder(**.idea/**) that's usually configuration files when you work on Jetbrains IDEs.

rajkumaar23:
@kartik918 Can you please review this too ?

kartik918:
This feature is a good one. 
Can you provide more descriptive screenshots/gif showing redirecting being changed? 
@rajkumaar23

rajkumaar23:
> This feature is a good one.
> Can you provide more descriptive screenshots/gif showing redirecting being changed?
> @rajkumaar23

dev

rajkumaar23:
> Great work implementing react router for seamless links.

Thanks! But there's a problem. `gh-pages` by default doesn't support SPA and hence when an user visits our application directly at `http://domain.com/resume`, `gh-pages` or any other server will throw **404**.
There are two hacks to avoid this : 
1) On a VPS, a .htaccess file would solve this.
2) On gh-pages, we'll have to parse the URL from our 404.html and then send it back to our application's root with the URL as a parameter and now we can redirect to the desired `/resume`.

So my question is should this information be mentioned in the README.md ?

kartik918:
``.htaccess`` redirection  would reduce load speed of the resume component I think

kartik918:
Since we'll be building on gh pages, we cannot use react router as GitHub has a protected server.

Redirecting 404 is a bad hack to use.

If you can design a solution with ``.htaccess`` , that would be better.

rajkumaar23:
.htaccess can only be used when you have a private server and not on 
`gh-pages`. Also, this isn't a bug. Internal navigation will still work 
good.

rajkumaar23:
> Since we'll be building on gh pages, we cannot use react router as GitHub has a protected server.
> 
> Redirecting 404 is a bad hack to use.
> 
> If you can design a solution with `.htaccess` , that would be better.

.htaccess can only be used when you have a private server and not on
`gh-pages`. Also, this isn't a bug. Internal navigation will work as intended rhrough react-router, when hosted with ghpages. 
For example, you can check the resume link at https://rajkumaar.co.in ( built with React and hosted by gh-pages) :)

saadpasta:
@rajkumaar23 I guess parsing the URL from our 404.html and then redirect it the resume page is the only hack we can do and I guess you have done it in the same way in your portfolio. 

@kartik918 what do you say because if we want this feature we have to do this hack to avoid the 404 page

rajkumaar23:
> @rajkumaar23 I guess parsing the URL from our 404.html and then redirect it the resume page is the only hack we can do and I guess you have done it in the same way in your portfolio.
> 
> @kartik918 what do you say because if we want this feature we have to do this hack to avoid the 404 page
I don't think redirection is required, because it's less frequent that the direct URL is hit. Hence, we can leave it to the user to follow the 404 hack or not. We can, may be, add a note on this in the README.md

rajkumaar23:
It is anyways mentioned [here](https://create-react-app.dev/docs/deployment/#notes-on-client-side-routing) (the page which you've mentioned in the deployment section of README.md

saadpasta:
@rajkumaar23 I guess you are right it's less frequent but I guess if someone will copy the URL and send it someone so he can view his CV/Resume then it leaves a bad impression.
We have to do the redirection hack I guess. if you could implement that it would be great.

rajkumaar23:
> @rajkumaar23 I guess you are right it's less frequent but I guess if someone will copy the URL and send it someone so he can view his CV/Resume then it leaves a bad impression.
> We have to do the redirection hack I guess. if you could implement that it would be great.

Sure I can. But the `404.html` has to be from the user's side. It won't be generated in the build dir.

rajkumaar23:
@saadpasta Do you want me to add `.htaccess` too ? For those who would love deploying on their own servers ?

kartik918:
@saadpasta In the end it's weighing if the feature is worth the redirection and addition of react-router dependencies just to have seamless link to resume. 
It's your decision, I'm not for it.

rajkumaar23:
> @saadpasta In the end it's weighing if the feature is worth the redirection and addition of react-router dependencies just to have seamless link to resume.
> It's your decision, I'm not for it.

I feel it looks professional when the resume loads up within your page, than from GDrive :)

kartik918:
> I feel it looks professional when the resume loads up within your page, than from GDrive :)

Redirecting 404 to serve our purpose isn't exactly professional practice in my opinion.

rajkumaar23:
> > I feel it looks professional when the resume loads up within your page, than from GDrive :)
> 
> Redirecting 404 to serve our purpose isn't exactly professional practice in my opinion.

Unfortunately, that's how SPAs work if **not** hosted on private servers :(

saadpasta:
> @saadpasta In the end it's weighing if the feature is worth the redirection and addition of react-router dependencies just to have seamless link to resume.
> It's your decision, I'm not for it.

@kartik918 I also think adding dependencies and a hack just for resume is not a good option and we will only add new features if we both agree on it. 
@rajkumaar23 let's keep this feature for the future 
Thanks for your contribution ❤️

PR #57, opened 2020-03-04
Fix website not secure

### LINKED ISSUE #56: Website Not Secure
While opening the website the padlock symbol doesn't show and on clicking it says parts of website is not secure. Better to change all http to https .

saadpasta:
Currently, our website is not secure because we are loading image from HTTP to fix this I have a loaded an image from https.
close #56

### DIFF
--- a/src/portfolio.js
+++ b/src/portfolio.js
@@ -104,7 +104,7 @@ const bigProjects = {
       link: "http://saayahealth.com/"
     },
     {
-      image: "http://nextu.se/wp-content/themes/nextu/images/og_img.jpg",
+      image: "https://drive.google.com/uc?id=1MXoXcQRK-pH8J82wyjCW4SJk5AxJe7tf",
       link: "http://nextu.se/"
     }
   ]


kartik918:
Suggestion :  We could serve the assets locally than over a CDN or GDrive link in this case for two reasons.

* App functions offline only too.

* The browser has to perform a handshake request to the remote server (in this case the GDrive).
Once the handshake is successful, the browser performs a second request to the remote server to get the files which can slow load times on bigger apps although almost non existent in this case.

Just something that might help in developing future apps. Cheers!

saadpasta:
> Suggestion : We could serve the assets locally than over a CDN or GDrive link in this case for two reasons.
> 
> * App functions offline only too.
> * The browser has to perform a handshake request to the remote server (in this case the GDrive).
>   Once the handshake is successful, the browser performs a second request to the remote server to get the files which can slow load times on bigger apps although almost non existent in this case.
> 
> Just something that might help in developing future apps. Cheers!

I did this url becuase it would be easier for others if they want to put pictures.
Will remember it for the future

Issue #39, opened 2020-01-02
Add new progress bar component


kartik918:
Since this repo is gaining popularity, I think we should create more opportunities to work or contribute for new people who might be interested. What do you think @saadpasta ?

saadpasta:
@kartik918 sound's great. But how can we achieve this?

kartik918:
I was thinking something like adding more features but still keeping them optional for people who don't want them in their portfolio.

Example, that progress bar thingy for showing your proficiency in a language or technology, we could work on it as a feature. I don't like that so I wouldn't want that but for people who do, it'd make the portfolio better for them. 
Also this would make the portfolio unique for anyone who uses it.

saadpasta:
@kartik918  Sounds Great. But I guess first we have make our deployment process easy and provide complete documentation on that. How can user deploy its site on Github Pages I have received lots of comments on this part

kartik918:
Its very easy but node package manager must be installed to deploy on Github pages.
Look it up [here](https://reactgo.com/deploy-react-app-github-pages/) and write a documentation accordingly.

saadpasta:
@kartik918 how can we achieve that optional thing do you have any idea.

saadpasta:
> Its very easy but node package manager must be installed to deploy on Github pages.
> Look it up [here](https://reactgo.com/deploy-react-app-github-pages/) and write a documentation accordingly.

Ok Let's see

kartik918:
> 
> 
> @kartik918 how can we achieve that optional thing do you have any idea.

I was thinking since all the components are individual, making them responsive and flexible enough to be able to work on their own would allow a user to make their own style of portfolio by removing the components they don't want.

i.e deleting the component files before using ``npm run deploy`` . 
Open to better solutions on this.

saadpasta:
@kartik918 That's a great idea and we can keep it for the future I think first foremost important work is to add a contact form. What do you say?

kartik918:
Contact "forms" are old and barely used, anyone who wants to contact anyone would email or use socials for quicker replies.
In my opinion we should work on a progress bar section for proficiency in skills first since that's a core of any portfolio (although I personally dislike them).

saadpasta:
@kartik918 I have asked some people and recruiters they say that if they have opened someone portfolio and want to contact them then contact forms is the best way to contact them they don't have to email them personally or call them they can avoid that hassle. 

[...]

PR #70, opened 2020-03-21
Add Progress Bar Component

kartik918:
Added a new progress bar component as was discussed in #39.
Shows user proficiency in certain skills.

Screenshot:

Screenshot (60)


naveen521kk:
@kartik918 Have you added animation while the progress bar loads? The colour should fill when the user views that component.

kartik918:
Addressed your feedback @saadpasta , now it appears if key is set to true.

kartik918:
Naveen I have a plan for overhauling the app to have animations for every component to make it more interactive. Adding grow transitions right now in an app that is completely static in terms of transitions would make it look off. 
If you want to help out in making the app interactive, I'll help you get started in the right direction.

naveen521kk:
@kartik918 I think that can be done but later. For now I am planning to make the Blog Component have latest posts from the RSS feed. Maybe you can add that as an Issue and lets solve it later.

PR #93, opened 2020-04-22
Security enhancment by using Environment Variables

IcyTv:
This is a proposal to change the token location to the environment variables, in order to enhance security and no leak any private keys in github repositories. Technically this could also make the btoa and atob conversions obsolete (Those are **NOT** security measures to protect private keys), but I didn't change that, because I first wanted to see if you even wanted to implement these changes


IcyTv:
I also realize now, that the token is exposed in the website itself.... Maybe this is not a significant security enhancement...

saadpasta:
@IcyTv Thanks for the PR. As we have clearly mentioned in the docs that you should select no scope while generating token. So a simple token just for querying public info

IcyTv:
Okay

PR #102, opened 2020-04-27
Update : GraphQl Query from pinnedRepositories to pinnedItems (Important)

saadpasta:
PinnedRepositories has been deprecated and has been removed from graphql query of Github. So currently we can't get pinned repo's from Github. I came to when I opened `saadpasta.github.io` and can't view my GitHub Projects.

The [changelog](https://developer.github.com/v4/changelog/2019-03-30-schema-changes/) for GitHub GraphQL schema (2019-03-30) mentions that the pinnedRepositories will be deprecated or removed by 2019-07-01. GraphQL API users are requested to use ProfileOwner.pinneditems instead. 

This has been fixed and we can view this API call on `saadpasta.github.io` it has been fixed there. 

#### Query (Can be tested on https://developer.github.com/v4/explorer/ )
`{
  user(login: "saadpasta") {
    pinnedItems(first: 6, types: [REPOSITORY]) {
      totalCount
      edges {
        node {
          ... on Repository {
            name
            description
            forkCount
            stargazers {
              totalCount
            }
            url
            id
            diskUsage
            primaryLanguage {
              name
              color
            }
          }
        }
      }
    }
  }
}
`


saadpasta:
This problem will be with everybody using DeveloperFolio Open Source Section

kartik918:
Oh nevermind, you mentioned it above. 
We can tell them manually to rebase their forks but that would take too long.

saadpasta:
@kartik918 should we create a release for this. 
I guess from now we should create releases for our updates what do you think.

kartik918:
I strongly agree with this. We could follow the [semantic versioning](https://semver.org/) syntax standard for releases.
We can update releases whenever there's a critical or major update.

PR #92, opened 2020-04-22
Added Dark Mode Feature

MuhammadHasham23:
I have added dark mode feature with a radio button on header.

The dark mode can be enabled and disabled as per user's choice. I've made a `style` object which changes the `background: black` and `color: white`. If `isDark` is set to `true`. Else the style won't be applied.

Since the header styles do not obey the styles injected on the parent `div`. We needed to pass the prop ( a function to trigger the change) down to `ToggleSwitch`. Which is the grand child of the parent component. So instead of passing down props through `Header`. Context API has been implemented with a `Provider` and `Consumer`. 

As it can be seen below,

image

A new directory for contexts is made to host contexts in future as well, currently it hosts `styleContext`

PR #104, opened 2020-04-29
Some changes to twitter component

naveen521kk:
I have done some changes to twitter component to make it look fine with other elements. I have made it transparent and removed the default header and added a new one. Check out the screenshot below.
tweets
Further, I did an edit changes talks from upper case to lower case.

### DIFF
--- a/src/containers/twitter-embed/twitter.css
+++ b/src/containers/twitter-embed/twitter.css
@@ -2,19 +2,35 @@
     display: block;
     margin: 0 auto;
     padding: 1.5rem;
+    box-shadow: rgba(0, 0, 0, 0.2) 0px 10px 30px -15px;
 }
-.tw-main-div {
-  margin-left: 15px; 
-  margin-right : 15px;
-  margin-bottom: 15px;
-  width: auto;
-  padding: 1.5rem;
+.centerContent:hover {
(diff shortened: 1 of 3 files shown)


kartik918:
Can you send screenshot of what changed?

naveen521kk:
This is the old one.
tweet1
This the new one.
tweets

naveen521kk:
And this PR is to help #92 .

kartik918:
I think the original header looks better and more fluent. What do you think @saadpasta ?

saadpasta:
> I think the original header looks better and more fluent. What do you think @saadpasta ?

@kartik918 I also think so. We could also make the component little small.

naveen521kk:
> 
> 
> I think the original header looks better and more fluent. What do you think @saadpasta ?

@kartik918 The original headers are not aligned with other components. So I had to change it. Further, I made the component transparent for the dark mode to work well.

kartik918:
> @kartik918 The original headers are not aligned with other components. So I had to change it. Further, I made the component transparent for the dark mode to work well.

Since the only thing we need out of this PR is the transparent bg, I think @MohammadHasham can accommodate for it in his PR. 
We'll reopen this when it is required. 
Thanks!

Issue #100, opened 2020-04-25
Setup CI/CD and new deployment domain


kartik918:
Our project has come far from how it used to be with a lot of changes and new components.

To show all newly added components and design changes, I think we should create it as a dummy portfolio. To make it clear, let's use all the components in the demo deploy build.

This would:
- keep the demo updated with all new features.
- since most people like to see the demo before using a project (myself included), this will show all possibilities of our project.

(This would also mean that you would have to move your personal deployment of the project to a private or other repo)
Let me know what you think about this.

saadpasta:
@kartik918 currently I am using another repo for `saadpasta.github.io`, not developerFolio. 

And I agree with you we should create a demo project so anyone who need to see demo can view all the possible components with DeveloperFolio and we should integrate CI/CD so we manually don't have to make a build and deploy it every time. 
We Could use Github Pages for our Demo and `gh-pages` branch for our build and the URL for our demo would be `saadpasta.github.io/developerFolio`. ✌️
What do you think.

kartik918:
Sounds good to me, we could use ``master`` branch for all new commits and merge it to ``gh-pages`` when there a certain amount of new changes introduced to master. This way it won't have to update again and again for minor changes.

Setting up CI for this would be a great boost to workflow. 

P.S your current build has missing components so you might want to look into that.

naveen521kk:
@saadpasta I found that you have deployed this in https://developerfolio.js.org/ . Please check out the issue. Rebuild the site again as there was a issue with it.

saadpasta:
@naveen521kk looking into it

naveen521kk:
Can we user [versel](https://vercel.com/docs/v2/git-integrations/vercel-for-github) CI/CD. It has great features and is free. It deploys all PR automatically and would be useful. What do you think @saadpasta and @kartik918 ?

kartik918:
I agree, Versel is great for CD. I would suggest @saadpasta to take up this issue.

saadpasta:
@kartik918 Done.

naveen521kk:
@saadpasta I will make a PR to gh-pages branch with Adding all the components. Let's use it as our deploy instead of [https://saadpasta.github.io](https://saadpasta.github.io). And then we can use js.org domain.

naveen521kk:
[...]

PR #115, opened 2020-05-06
New Deployment for protofolio.

naveen521kk:
Fixing #100


naveen521kk:
@saadpasta After merging it I would make a PR in js.org for domain., with that we can close that issue.

PR #127, opened 2020-05-14
Remove Font-Awesome 

### LINKED ISSUE #126: Remove src/assests/font-awesome folder
The folder located at `src/assests/font-awesome` has many issues in Codacy and as well it is not *required*. I also *found* there is a link to font-awesome CSS in `index.html` and that would do. So instead of importing the whole font-awesome using a *CDN* is better.

naveen521kk:
Fixes #126 
- Removed `src/assests/font-awesome` folder
- Made changes to `src/index.js` regarding it
- Updated cdn link in `public/index.html` for *font-awesome*


naveen521kk:
Test deploy @https://mknaveen.netlify.com

kartik918:
Looks good to me except I see no icon for "toTop" button.

image

We should look into that.

public/index.html
```diff
@@ -38,7 +38,7 @@
     </script>
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
 
-    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/FortAwesome/Font-Awesome@5.13.0/css/all.min.css">
```
kartik918, line 41: Do we need to update ``fa`` CDN? If not, then let's keep the old one since this will increase file size download on load, a problem we are already dealing with.

kartik918:
Let's change the toTop button icon to ``chevron-up`` since it adds to the minimalist look. What do you think?

public/index.html
naveen521kk: In version 5 Font awesome had depreciated `fa` and instead `fas` is used. And the fact they renamed that icon 'hand-o-up' from to `hand-point-up` in version 5. Import that one script is enough.

naveen521kk:
I personally like the previous icon very much that it has not been used anywhere else and a *Uniqueness* for developerFolio. The icon `chevron-up` is nearly used on many websites.  Let's ask for others feedback also. So we can open it as an Issue.

kartik918:
@saadpasta, what do you think about the scroll up icon?

I think Chevron because it's minimal and it's smaller, fixing the screen space issue on smaller screens too. But I'm open to whatever the third vote says.

saadpasta:
@kartik918 I think the previous icon was really cool.

PR #136, opened 2020-05-21
Added Experiences Section

HenryHengZJ:
I have added the Experience Section.
The Experience section can be enabled or disabled via setting viewExperiences in portfolio.js.


naveen521kk:
It would good if font size is reduced on the screen size reduces. Further, doesn't look good on mobiles. Make it match other components.

src/containers/workExperiences/workExperiences.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import "./workExperiences.css";
+import ExperienceCard from "../../components/experienceCard/ExperienceCard";
+import { workExperiences } from "../../portfolio";
+import { Fade } from "react-reveal";
+
+
+export default function Experiences() {
```
kartik918, line 8: Let's keep the export name consistent to the file name for standards and less confusion.

Also rename file to `workExperience` and update wherever required in other files too.

src/portfolio.js
```diff
@@ -115,6 +115,36 @@ const techStack = {
...
+      companylogo: "https://www.referralcandy.com/wp-content/uploads/2017/06/quora-featured-image-2-740x740.png",
+      date: "May 2017 – May 2018",
+      descrip: "Write clean, maintainable code using best industry standards in Web languages, and integrate with existing APIs."
+    },
+    {
+      role: "Software Engineer Intern",  
+      company: "Airbnb",
+      companylogo: require("./assests/images/airbnbLogo.jpg"),
+      date: "Jan 2015 – Sep 2015",
+      descrip: "Develop and maintain new/current software applications, meeting user requirements and complying with technical specifications."
```
kartik918, line 143: Let's change `descrip` prop name to `desc` for consistency and readability. Also update accordingly in other files.
Switch `desc` to Lorem Ipsum

HenryHengZJ:
> It would be more appropriate to use placeholder information instead of accurate dates and details (something like Lorem Ipsum, we can stick to company icons or some other placeholder images to show the color extraction feature).
> 
> Also, it seems to be slightly inconsistent on smaller screens.
> 
> image
> 
> Love the card design BTW!

Thanks for the changes suggestions. I noticed the inconsistent look at iPhone5/SE, the same goes for Achievement and Certifications. Changes needed to be made there too.

image

kartik918:
Great find!

We should investigate what's causing this in the first place before moving to a fix.

HenryHengZJ:
The problem is at the CSS file of the grid divider:
`grid-template-columns: repeat(auto-fit, minmax(300px, 1fr));`
If you reduce 300px to 290px, it fixes the problem.

kartik918:
Even for larger screens? I don't want it to look slightly off from the centre. Thanks!

HenryHengZJ:
The problem only occurs when tested with Iphone5/SE, works fine on larger screens. Making it 290px will not shift the grid, the only thing changed is the min card's width is now 290px when screen size reduces. 
image

Below is the image when tested with Iphone5/SE. 

image

kartik918:
Having slightly smaller width isn't a "problem" as severe as having cards offset from the center. I think this looks great. Push the requested changes and we are good to go!

src/portfolio.js
HenryHengZJ: do you want to change the date to Lorem Ipum as well?

kartik918: Nah, just add some place holder date or let it be as it is right now.

README.md
```diff
@@ -21,6 +21,7 @@ If you created something awesome and want to contribute then feel free to open P
 ## Sections 
 ✔️ Summary and About me\
 ✔️ Skills \
+✔️ Experiences \
```
kartik918, line 24: Let's change this to "Work Experience", seems more appropriate.

kartik918: Nevermind, did it myself.

kartik918:
@saadpasta you can add your own work experience in a follow up PR or we can move to neutral filler information for the demo build. Whatever you decide on this, I'm for it.

naveen521kk:
You @kartik918 forgot to add @HenryHengZJ as a contributor.

kartik918:
@allcontributors please add @HenryHengZJ for code and design

HenryHengZJ:
Thanks guys for the help and suggestions!

PR #141, opened 2020-05-24
Fix readme contributors section

kartik918:


kartik918:
Fixes all contributors bot bringing us back to the contributors section.

PR #151, opened 2020-05-28
Update Github UserName 

saadpasta:
Add my Github Username to show my profile instead of undefined.

### DIFF
--- a/src/portfolio.js
+++ b/src/portfolio.js
@@ -150,7 +150,7 @@ To know how to get github key look at readme.md */
 
 const openSource = {
   githubConvertedToken: "Your Github Converted Token",
-  githubUserName: "Your Github Username",
+  githubUserName: "saadpasta", // Change to your github username to view your profile in Contact Section.
   showGithubProfile :"true" // Set true or false to show Contact profile using Github, defaults to false 
 };
 


naveen521kk:
I think we just dont change the one in main thing, instead we can create a new `portfolio.js` in `.github` folder and to copy it to `src/portfolio.js` while deploying. In `.github/portfolio.js` you could update with necessary details. @saadpasta

saadpasta:
@naveen521kk do you have any idea how can we do that

saadpasta:
We can create an issue for that and do it later. But for now, we have to do this because it better show something other than undefined

naveen521kk:
So initially, let's create a new file in `.github` folder with name as `portfolio.js`. We can edit all the changes we want to personalize our deploys like making all components accessible by adding you GitHub profile and repo access tokens. Change in `.github/workflow/node.js.yml` such that to copy the file `.github/portfolio.js` to `src/portfolio.js`. After that build, the command can be run and changes can be pushed to `gh-pages` by GitHub Actions. By this, we don't need to change anything in the master branch for deployment and would help the users also. @saadpasta

naveen521kk:
This is just _temporary_ solution. I will make the necessary changes in new PR. @saadpasta

saadpasta:
@naveen521kk Let's merge this and then create an issue for portfolio.js.

PR #152, opened 2020-05-28
Seperate CI with Other

naveen521kk:
I have separated CI with other things. And useful for deploying to master branch using Github Actions.

- Create a new `Portfolio.js` in `.github` folder.
- Use it while deploying instead of original one for all components to be displayed.

### DIFF
--- a/.github/portfolio.js
+++ b/.github/portfolio.js
@@ -149,8 +149,8 @@ const workExperiences = {
 To know how to get github key look at readme.md */
 
 const openSource = {
-  githubConvertedToken: "Your Github Converted Token",
-  githubUserName: "Your Github Username", // Change to your github username to view your profile in Contact Section.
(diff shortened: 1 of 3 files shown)


naveen521kk:
Set an *environment variable* called `REACT_APP_GITHUB_TOKEN` with a personal access token for Github Repo Card. @saadpasta .
Also @kartik918 please make sure you update `.github/portfolio.js` every time there is any change in `src/portofolio.js` for any PR, so that there is no error while deploying it.

kartik918:
This work was supposed to be handled by @saadpasta since it involved env variables and why does `.github/portfolio.js` needs to be updated manually? I'm not sure but that stuff should happen automatically.

naveen521kk:
Let @saadpasta take up this PR. So *closing* it.

PR #155, opened 2020-05-30
Add bullet points in experience description

### LINKED ISSUE #154: Add subPoints section in workExperiences section
First of all great work @saadpasta and all other contributors, really an awesome project.

## Summary
In the experience section, It would be great to have a functionality to add subpoints along with the general description.

### Motivation
Actually I was using the project to create my own portfolio, but some of my internships involved me working on several projects instead of one. The idea to keep all that detail in a single paragraph did not feel good, so I added a subPoints section to the workExperiences variable, and rendered a list using that.

### Additional context
portfolio1

AkshayCHD:
Fixes #154 
Allows user to add optional subpoints in the experience section


src/components/experienceCard/ExperienceCard.js
```diff
@@ -15,6 +15,10 @@ export default function ExperienceCard({ cardInfo }) {
     return typeof values === "undefined" ? null : "rgb(" + values.join(', ') + ")";
   }
 
+  const GetSubPoints = ({ subPoints }) => {
```
kartik918, line 18: Could you rename `subPoints` to something like `desc-bullet` everywhere. It would seem more appropriate with the other values.

src/containers/workExperience/WorkExperience.js
```diff
@@ -22,7 +22,8 @@ export default function WorkExperience() {
                                         desc: card.desc,
                                         date: card.date,
                                         companylogo: card.companylogo,
-                                        role: card.role
+                                        role: card.role,
+                                        subPoints: card.subPoints
```
kartik918, line 26: Same here about renaming the subPoints.

src/portfolio.js
```diff
@@ -126,21 +126,25 @@ const workExperiences = {
...
+      desc: "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.",
+      descBullets: [
+        "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
+        "Lorem ipsum dolor sit amet, consectetur adipiscing elit"
+      ]
     },
     {
       role: "Front-End Developer",   
       company: "Quora",
       companylogo: "https://www.referralcandy.com/wp-content/uploads/2017/06/quora-featured-image-2-740x740.png",
```
kartik918, line 138: Could you get the png and save it in the assets folder so the app serves it locally like the other two logos.

AkshayCHD: @kartik918 Done

kartik918:
Looking forward to more contributions from you! @allcontributors add @AkshayCHD for code.

AkshayCHD:
@kartik918 thanks :), Also, there was another issue, don't you think there should be an env file to manage one's github token, one would definitely not want to keep his token in his source code, we can keep that optional, should I raise an issue for that as well.

kartik918:
Thanks for that idea. We assumed that the person would be keeping the repo private or use the env variables themselves to deal with this but you have a good point here.

I'm wondering if gh secrets can be used in a normal repository that isn't an action. The only experience I've had it with it was when I was designing a gh action.

We would also need to update the README to guide users on setting up the env file.

AkshayCHD:
Sure will do that as well, let me create an issue for that

Issue #156, opened 2020-05-31
Add "outward" transitions for button hover effects


kartik918:
Current button hover animations work smoothly on the first half and return to their initial state instantly on the other half of the animation (i.e taking cursor off the buttons)

Adding complementary CSS `transition: ease-in 0.3s;` to the button component `.main-button` class should work.

sparsh-99:
Hey @kartik918, what you actually want with CSS hover?
Can you please give me a brief description?
It will be very helpful in contributing.
Thanks

kartik918:
I'm sorry if I wasn't clear enough in the description, basically what Saad mentioned in your [PR](https://github.com/saadpasta/developerFolio/pull/179#pullrequestreview-427376605). 
Check the link to his deployment of the portfolio and update the CSS transitions with it, adding `transform: translateY(-3px);` would be a plus.
Let me know if you need more information on this @sparsh-99 .

PR #179, opened 2020-06-08
Update main-button css

sparsh-99:
#156 
### New Features:-
1. Added the animations in the .main-button class
2. Updated the buttons to look it more attractive.

### DIFF
--- a/src/components/button/Button.css
+++ b/src/components/button/Button.css
@@ -18,31 +18,8 @@
 .main-button:hover {
   background-color: #ffffff;
   color: black;
-  transition: ease-in 0.5s;
-}
-.main-button span {
-  cursor: pointer;
-  display: inline-block;
-  position: relative;
-  transition: 0.5s;
-}
-
-.main-button span:after {
-  content: '\00bb';
-  position: absolute;
-  opacity: 0;
-  top: 0;
-  right: -20px;
-  transition: 0.5s;
-}
-
-.main-button:hover span {
-  padding-right: 25px;
-}
-
-.main-button:hover span:after {
-  opacity: 1;
-  right: 0;
+  transition: ease-in 0.3s;
+  transform: translateY(-3px);
 }
 .project-button{
   display: flex;
--- a/src/components/button/Button.css
+++ b/src/components/button/Button.css
@@ -18,7 +18,31 @@
 .main-button:hover {
   background-color: #ffffff;
   color: black;
-  transition: ease-in 0.3s;
+  transition: ease-in 0.5s;
+}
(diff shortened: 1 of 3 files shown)


src/components/button/Button.css
```diff
@@ -18,7 +18,31 @@
 .main-button:hover {
   background-color: #ffffff;
   color: black;
-  transition: ease-in 0.3s;
+  transition: ease-in 0.5s;
```
kartik918, line 21: I see you've increased the transition time, we don't need slower animations so let's revert this.

sparsh-99: Ok, thanks. Working on the outward animation
Also, I will revert the transition time.
Thanks

sparsh-99:
Sorry @kartik918 @saadpasta , by mistake, I commit a new file here which has no reference here and I am unable to revert it back? Can you please help or delete that delete. It will be very helpful. Thank You

saadpasta:
@sparsh-99 `git revert 7d6b832` should work I guess.

sparsh-99:
Thank you @saadpasta , but due to some internal errors, the previous log was deleted. Now `git log` is not showing this commit. That's why I can't revert it back

src/components/githubRepoCard/GithubRepoCard.js
```diff
@@ -13,7 +13,7 @@ export default function GithubRepoCard({ repo }) {
     <div>
       <div className="repo-card-div" key={repo.node.id} onClick={() => openRepoinNewTab(repo.node.url)}>
         <div className="repo-name-div">
-          <svg aria-hidden="true" className="octicon" height="16" role="img" viewBox="0 0 12 16" width="12" className="repo-svg">
+          <svg aria-hidden="true" className="octicon" height="20" role="img" viewBox="0 0 12 16" width="14" className="repo-svg">
```
kartik918, line 16: Remove changes to this file.

src/containers/projects/Project.css
```diff
@@ -1,5 +1,5 @@
 .project-title {
-  font-size: 56px;
+  font-size: 52x;
```
kartik918, line 2: Discard this file from the PR too.

sparsh-99:
Hey @kartik918 @saadpasta , What if I again opened the PR for the same issue as reverting the last commit got some issues??
Thank You.

kartik918:
If you would like to then it's fine, just do everything in one commit and reference the issue in the PR desc @sparsh-99

sparsh-99:
Sure @kartik918 . Thank You.

PR #181, opened 2020-06-09
Add Loading screen

### LINKED ISSUE #175: Add loading Screen to Twitter Component, Github Profile and Github Repo
## Summary
Need to add a loading Screen in Twitter Component, Github Profile and Github Repo just because they are taking time to load and leaves a bad impression on first look. A simple Loading Screen would do. 

### What's new?
Adding loading Screen.

### Motivation
Just come up with it while visiting the site.

naveen521kk:
Fixes #175 
In this PR,
- Adding a New Loading COmponent
- Used the loading component in GitHub Repo, GitHub Profile and Twitter Component
- Used [SpinKit](https://tobiasahlin.com/spinkit/) for loading Screen.
- Fixed Error Loading messages in GitHub Repo, GitHub Profile and Twitter Component.


src/App.css
```diff
@@ -27,4 +27,4 @@
 
 .subTitle {
   color: #868e96;
-}
+}
```
kartik918, line 30: Let's revert the changes in this file.

src/containers/loading/loading.css
```diff
@@ -0,0 +1,121 @@
...
+
+@keyframes sk-circleBounceDelay {
+  0%, 80%, 100% {
+    -webkit-transform: scale(0);
+            transform: scale(0);
+  } 40% {
+    -webkit-transform: scale(1);
+            transform: scale(1);
+  }
+}
```
kartik918, line 121: Let's add another line here.

src/containers/profile/Profile.js
```diff
@@ -1,10 +1,11 @@
...
-import GithubProfileCard from "../../components/githubProfileCard/GithubProfileCard";
+import React, { useState, useEffect ,lazy, Suspense } from "react";
 import axios from "axios";
 import { openSource } from "../../portfolio";
 import Contact from "../contact/Contact";
+import Loading from "../loading/Loading";
 
-
+const renderLoader = () => <Loading />;
+const GithubProfileCard = lazy(() => import('../../components/githubProfileCard/GithubProfileCard'));
```
kartik918, line 8: Good use of lazy loading here, nice!

src/containers/projects/Projects.js
```diff
@@ -1,13 +1,16 @@
...
-import GithubRepoCard from "../../components/githubRepoCard/GithubRepoCard";
 import Button from "../../components/button/Button";
+import Loading from "../loading/Loading";
 import { openSource, socialMediaLinks } from "../../portfolio";
-import { Fade } from "react-reveal";
+
 
 export default function Projects() {
+  const GithubRepoCard = lazy(() => import('../../components/githubRepoCard/GithubRepoCard'));
+  const FailedLoading = () => <div className="centerContent"><h3>Looks like you have not configured for Github Repo.</h3></div>;
```
kartik918, line 12: We don't have to force the user to show his gh projects, lets return null if not found, and to help debug, we can log "Projects section not configured" or something.

src/containers/projects/Projects.js
```diff
@@ -78,6 +86,9 @@ export default function Projects() {
         </div>
         <Button text={"More Projects"} className="project-button" href={socialMediaLinks.github} newTab={true} />
       </div>
-    </Fade>
```
kartik918, line 81: Does it work if we put `Fade` inside Suspense, if yes then let's keep it, in earlier files too.

saadpasta:
@kartik918 @naveen521kk  I don't think we need suspense for this we could simply use the state to show and hide loader what do you suggest.

And can we use Svg for Loader something [like](https://github.com/saadpasta/react-blog-github/blob/master/src/Components/Common/LoaderSvg.js) this in react-blog-github. what do you guys think?

kartik918:
SVG for loader is a preference I think so it doesn't matter, using the spinkit looks nice too and is just simple CSS which is awesome. I'm okay either ways so I'll wait for Naveen's opinion on it.

Agree with avoiding to use suspense for now.

naveen521kk:
So I would change using suspense. Also, I will CSS for the loading screen as SVG won't be that good imo. Maybe I will change the colour to our signature purple.

naveen521kk:
@kartik918 @kartik918 Looks like for using `React.lazy` using `Suspense` is a must and also is available in a stable version. See this [doc](https://reactjs.org/docs/code-splitting.html#reactlazy). Also using Suspense for data fetching is what is an experimental version. Check out this [doc](https://reactjs.org/docs/concurrent-mode-suspense.html). So now you guys should decide whether to use lazy loading for GitHub Repo, GitHub Profile and Twitter Component or not. Maybe if not, adding a loading screen can easily be done with `useState`. Also, there is quite a lot of advantage of using lazy loading.

kartik918:
Sorry, I was kinda outdated on my knowledge about Suspense and skimmed through the docs on it before making my review. The points you make are valid and the prototype build was working fine so I don't see a problem with it. My only concern is, stuff in experimental phase in React can (and almost always does) change so it isn't something to rely on for production builds. Hopefully it doesn't mess up our builds in a future update down the road.

@saadpasta can take a final decision regarding this, I'm again okay either ways.

saadpasta:
@naveen521kk @kartik918 If you guys are okay with so let's go with Suspense and lazy loading so let's go with it.

naveen521kk:
So, alongside the review is there any changes I should do?

kartik918:
Nothing I can think of right now @naveen521kk , you can add something if you feel like it but let us know!

src/containers/projects/Projects.js
naveen521kk: Well adding `Fade` doesn't show the loading screen, so just leaving it as is.

naveen521kk:
Well, I really don't know why Versel is failing. It is perfect in my local environment. @saadpasta Could you please check out as it is asking to log in if needed to check build.

PR #184, opened 2020-06-12
Remove : Axios and Replace API call with GraphQl

### LINKED ISSUE #182: Remove axios to fetch user data from Github
## Summary
Currently, we are using Axios to fetch user Github Profile to view its contact section. The API call is returning a lot of unwanted data that we don't need.

### What's new?
We should remove Axios and use Github GraphQl V4 API to fetch user profile from Github using the below schema. You can test this on [GraphQl Api Explorer](https://developer.github.com/v4/explorer/)

```js
query { 
  user(login:"saadpasta") { 
    name
    bio
    isHireable
    avatarUrl
    location
  }
}
```

### Motivation
This would remove one of our dependencies and will maintain the same thing across the whole portfolio. It would also remove the unnecessary data that we are getting from the v3 API call

### Currently Used Code. (That should be replaced with the above GraphQL call)

```js
function getProfileData() {
    const client=axios.get("https://api.github.com/users/"+openSource.githubUserName)
      .then(function(response){
            setProfileFunction(response.data);
    });
  }
```

saadpasta:
fixes #182

### DIFF
--- a/src/containers/profile/Profile.js
+++ b/src/containers/profile/Profile.js
@@ -1,7 +1,6 @@
 import React, { useState, useEffect } from "react";
 import GithubProfileCard from "../../components/githubProfileCard/GithubProfileCard";
-import ApolloClient from "apollo-boost";
-import { gql } from "apollo-boost";
+import ApolloClient, { gql } from "apollo-boost";
 import { openSource } from "../../portfolio";
 import Contact from "../contact/Contact";
 
@@ -41,11 +40,9 @@ export default function Profile() {
       });
   }
   useEffect(() => {
-
     if (openSource.showGithubProfile === "true") {
       getProfileData();
     }
-    
   }, []);
   if (openSource.showGithubProfile === "true") {
     return <GithubProfileCard prof={prof} key={prof.id} />;
--- a/src/containers/profile/Profile.js
+++ b/src/containers/profile/Profile.js
@@ -1,9 +1,9 @@
 import React, { useState, useEffect } from "react";
 import GithubProfileCard from "../../components/githubProfileCard/GithubProfileCard";
-import axios from "axios";
+import ApolloClient from "apollo-boost";
+import { gql } from "apollo-boost";
 import { openSource } from "../../portfolio";
 import Contact from "../contact/Contact";
-
 
 export default function Profile() {
   const [prof, setrepo] = useState([]);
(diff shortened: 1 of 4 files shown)


src/containers/profile/Profile.js
```diff
@@ -1,27 +1,55 @@
...
+      `,
+      })
+      .then((result) => {
+        setProfileFunction(result.data.user);
+      });
   }
   useEffect(() => {
-    getProfileData();
+
+    if (openSource.showGithubProfile === "true") {
```
naveen521kk, line 45: Do we really need to check this condition?

src/containers/profile/Profile.js
```diff
@@ -1,27 +1,55 @@
...
+          isHireable
+          avatarUrl
+          location
+        }
+    }
+      `,
+      })
+      .then((result) => {
+        setProfileFunction(result.data.user);
+      });
```
naveen521kk, line 41: Add an error handler here.

src/containers/profile/Profile.js
```diff
@@ -1,27 +1,55 @@
 import React, { useState, useEffect } from "react";
 import GithubProfileCard from "../../components/githubProfileCard/GithubProfileCard";
-import axios from "axios";
+import ApolloClient from "apollo-boost";
```
naveen521kk, line 3: Instead of two import statement from same library make it as a single one. See the suggested changes.
```suggestion
import ApolloClient, { gql } from "apollo-boost";
```

src/containers/profile/Profile.js
```diff
@@ -1,27 +1,55 @@
...
+import { gql } from "apollo-boost";
```
naveen521kk, line 4:
```suggestion

```

src/containers/profile/Profile.js
saadpasta: Yeah if someone has set false in showGithubProfile to false we won't call API

naveen521kk: We, already have a condition check-in line 2
https://github.com/saadpasta/developerFolio/blob/772d4cc9814a97db9b67044581868b240d0f21de/src/containers/profile/Profile.js#L22

saadpasta: That's for the component not for the API call

naveen521kk: It would be good to log this in the console.

src/containers/profile/Profile.js
naveen521kk: Basically add these lines here
```suggestion
      }
      .catch(function (error) {
        console.log(error);
      };
```

PR #186, opened 2020-06-14
preloaders for portfolio

samridhi18:
Different style of loaders could be used for pre-loading styles in portfolio


naveen521kk:
I don't understand what is this PR indented to do?

kartik918:
@samridhi18 avoid opening PRs to a project with no prior information to the maintainers. This would be something that isn't required or not intended and then your work would be wasted which isn't something you would like.
The right SOP would be to open an issue for it to start a discussion first and then work if it's agreed upon.
Thanks!

Issue #193, opened 2020-06-27
Experience cards don't display properly on mobile browser


mcsheha:
On mobile browser, experience headings and logos stack at top of section.  

**Screenshots**

![Screenshot 2020-06-26 at 8 11 07 PM](https://user-images.githubusercontent.com/21299027/85909812-b3cc6000-b7e9-11ea-88a9-7ef955415693.png)

**Smartphone (please complete the following information):**
 - Device: iPhone Xs Max
 - OS: iOS 13
 - Browser: Safari, Chrome

**Additional context**

tiagodenoronha:
Same issue happens on Mac. See attached screenshot:

<img width="1606" alt="image" src="https://user-images.githubusercontent.com/27889458/86272109-f6b17d80-bbc5-11ea-8937-c9a4a4c90473.png">

naveen521kk:
This is disturbing. Looks like a weird bug and I doubt `color theif` is the culprit. Maybe, I will try using lazy like I did for other sections if that works I will make a PR tomorrow if not I prefer reverting that Commit which added it. I need your comments on this @kartik918 and @saadpasta

saadpasta:
@naveen521kk I did not look into this much I think you are right about **color theif**. Try fixing this if possible otherwise, just remove the library and a key for the background color for now

kartik918:
We certainly won't and shouldn't revert breaking commits. Instead we can look into the cause and patching it ASAP.

Still a weird bug because I remember testing for responsiveness issues. I doubt shifting to suspense would fix it. Regardless, we should look into this and open a PR fixing it. 

I'll check it in depth later and share what I find

kartik918:
I checked on FF and Chrome, even on mobile to confirm but I can't reproduce the bug. It looks fine on smaller screens for me,

![Screen Shot](https://user-images.githubusercontent.com/48270786/86324208-d6340280-bc5b-11ea-93ba-c60d4796eb5c.png)

My theory judging from the screenshots above, it is caused by images not loading and `color-thief` complaining because it can't extract colors, thus the empty grey area above the cards.

This looks like a Safari only issue, which shouldn't happen ideally, @mcsheha could you confirm and provide some steps to reproduce this?

mcsheha:
@kartik918 This occurs for me in iOS 13 in both Safari and Chrome. I disabled content/ad blockers and tried various networks including cellular and wifi and the problem persists.

naveen521kk:
[...]

PR #201, opened 2020-07-03
Fixed experience cards issue on safari

abhishekashyap:
Fixed the issue in accordance with #193  that caused experience cards to not display correctly on safari on iOS and macOS.

Screenshot 2020-07-03 at 3 09 22 PM

Screenshot 2020-07-03 at 3 09 05 PM


kartik918:
@allcontributors add @abhishekashyap for bug fixes!

Issue #174, opened 2020-06-07
Use Headroom for header


kartikcho:
## Summary
Let's add [Headroom](https://github.com/KyleAMathews/react-headroom) to use for the header.

### What's new?
Header hides while scrolling but is stickied to top when you scroll back up.

### Motivation
H E A D E R

### Additional context
This is a good and simple issue for any first time contributors but we should add this if it doesn't get taken up in a week or two.

Also, if you're working on the PR, add it to the list of dependencies in the README and link the original repo to credit the author.

sparsh-99:
I want to work on the issue, @kartik918

kartik918:
Okay, this issue is your now @sparsh-99

sparsh-99:
@kartik918 , for headroom tag, I want to change HTML file. Can you please tell me where all HTML work is saved so that I can review it for the change?

naveen521kk:
@sparsh-99 Any updates?

sparsh-99:
@naveen521kk I started working on it. A PR will be opened within a few days.

naveen521kk:
@kartik918 Lets wait for few more days for any new contributor. If not, we could do it.

kartik918:
Sure! You could invite someone you know to take this up as their first issue. You can mentor them through the issue and possibly grow a new contributor!
Or we could just finish it in a bit. Right now the dark mode PR is becoming a blocker for all existing PRs which needs more attention I think.

archit0210:
I will work on this issue @kartik918

archit0210:
@kartik918 need help!

kartik918:
What's the problem?

archit0210:
@kartik918 as i was just going through the workflow i want to know that you want changes in all pages header or only front page?

kartik918:
The portfolio is a single page application and header is a separate component.

npm install the headroom dependency I mentioned in the issue and wrap the header component inside it. Look at their docs for more help.

This should help you out!

archit0210:
![Capture](https://user-images.githubusercontent.com/67779397/86571972-e9aed880-bf8f-11ea-8e96-9555df8c32ce.PNG)
@kartik918 i tried to do fix the header but he thing is the header is transparent so the text is there(behind the header)

kartik918:
You can add necessary CSS to header's background.

Also why is the header so "low" from the top?

archit0210:
that's the snapshot after scrolling

archit0210:
[...]

PR #205, opened 2020-07-08
Headroom added

archit0210:


src/components/header/Header.js
```diff
@@ -2,12 +2,15 @@ import React from "react";
...
 import {greeting, workExperiences} from "../../portfolio";
+import { headroom } from "headroom";
 
 function Header() {
   const exp = workExperiences.viewExperiences;
   return (
-    <Fade top duration={1000} distance="20px">
-    <div>
+  <Fade top duration={1000} distance="20px">
+    <script src="https://unpkg.com/headroom.js"></script>
```
naveen521kk, line 11: You should not use script tag here.

src/components/header/Header.js
```diff
@@ -48,6 +51,9 @@ function Header() {
         </ul>
       </header>
     </div>
+    <script>
```
naveen521kk, line 54: Again this also.

src/components/header/Header.js
naveen521kk: Instead here
```html
<Headroom>
```
Should be coming.

archit0210: but the import headroom from headroom is not working

kartik918: I tried to guide you in my comment on #174 too, take a look at Headroom's documentation.

archit0210: i added and followed all the steps and this error occurred
Capture

kartik918: Change Headroom.js to `"Headroom"` @archit0210

archit0210: i did, there is no change in error

archit0210: okay, now is this correct?

archit0210: okay, now is this correct?

kartik918: Umm, why are you initializing Headroom, I don't remember exactly but it isn't as complicated as your implementation, you just need to import Headroom and wrap the Header component in it. 
Headroom will make the header sticky itself!

archit0210: oh sorry i forgot, i have wrapped header component...

archit0210: i tried restarting and checked the names of files but couldn't resolve!!

kartikchou:
Hey @archit0210 I see you're struggling with this so I'm gonna take this issue up and open a PR for it, maybe it'll help clear much of your confusion. Is that okay?

src/components/header/Header.js
archit0210: No

kartikcho:
Superseded by #209 , @archit0210 you might want to take a look!

PR #209, opened 2020-07-12
Add headroom, add yarn.lock

### LINKED ISSUE #174: Use Headroom for header

kartikcho:
Resolves #174 and supersedes #205 

Increased width of header to 100% and added shadow 
image

Let me know if the shadow looks good.


kartikcho:
I agree too partially but removing it looks kinda out of place for a header but I'm okay whatever is decided.

naveen521kk:
I don't like the shadow as it looks little off from theme.

kartikcho:
Done @naveen521kk !

src/components/header/Header.js
```diff
@@ -1,13 +1,13 @@
 import React from "react";
+import Headroom from "react-headroom";
 import "./Header.css";
 import {Fade} from "react-reveal";
```
naveen521kk, line 4: This import is not required.
```suggestion
```

naveen521kk:
Also, @kartikcho add the React Headroom to the Readme file as well.

src/components/header/Header.js
kartikcho: oh, thanks!

PR #222, opened 2020-08-06
Update link to Open Source Projects in Readme

### LINKED ISSUE #218: Open Source Project section is missing
I am facing problem with Open source project section and github user data, might be related with #101 
Issue is present in both development and final build
check the live page  ([here](https://flyingbackdoor.github.io))

I am getting the below warnings in log
```
./src/containers/projects/Projects.js
  Line 18:6:  React Hook useEffect has a missing dependency: 'getRepoData'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```
```
./src/containers/profile/Profile.js
  Line 54:6:  React Hook useEffect has a missing dependency: 'getProfileData'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```

When I deployed for the first time Github revoked my access token even after adding **.env** file in gitignore. So I used the old method, converting it and directly assigning to variable which worked. But it is still not fetching any details.

Code Preview👇
image

naveen521kk:
This is much-awaited documentation.
Fixes #218

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,37 +73,7 @@ $ npm start
 ```
 ## Github Setup For Open Source Projects
 
-Generate a Github personal access token using these [Instructions](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line) `Make sure you don't select any scope just generate a simple token`
-
-1. Create a file called .env in the root directory of your project, check the base file
-
-Note: Instead of creating a .env file, you can just run this command "cp env.example .env" inside the root directory 
-
-```bash
-- DeveloperFolio
-  - node_modules
-  - public
-  - src
-  - .env         <-- create it here
-  - env.example  <-- this is the base file
-  - .gitignore
-  - package-lock.json
-  - package.json
-```
-
-2. Inside the .env file, add key `REACT_APP_GITHUB_TOKEN` and assign your github token like this.
-
-```javascript
- // .env
-
-  REACT_APP_GITHUB_TOKEN = "YOUR GITHUB TOKEN HERE"
-
-```
-
-Set `showGithubProfile` to true or false to show Contact Profile using Github, defaults to false.
-
-Note: Open Source Projects section only show pinned itens of your Github. If you are seeing [this](https://i.imgur.com/Hj6mu1K.png), please follow this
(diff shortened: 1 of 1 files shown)


naveen521kk:
The one in the wiki works perfectly, and also you would be doing the same for vercel. 

The dotenv AFAIK didn't work for anyone including me. On searching, I found we didn't configure it properly. We would need to add the dependency https://www.npmjs.com/package/dotenv , and configure it, if I'm not wrong. 

Maybe, this weekend I will work making it work in that method.

saadpasta:
@naveen521kk It's really weird It's working perfectly for me. I guess it must working for you on local and then when put it on dev the token would be revoked. Is that correct

naveen521kk:
`.env` method is working perfectly for me locally.
I haven't pushed it anywhere. But I don't how it doesn't work for others.

saadpasta:
@naveen521kk I think we should keep both the solutions.

kartikcho:
> 
> 
> I think we should keep both the solutions.

Same, people in my connections are using this portfolio but had no troubles with the setup. It's still pretty weird seeing users run into that problem.

naveen521kk:
Maybe, let's say, if the above one fails, go see the wiki page for an alternate one? Will that do?

saadpasta:
> Maybe, let's say, if the above one fails, go see the wiki page for an alternate one? Will that do?

Sounds Great 😁

README.md
```diff
@@ -95,16 +96,16 @@ Note: Instead of creating a .env file, you can just run this command "cp env.exa
...
   REACT_APP_GITHUB_TOKEN = "YOUR GITHUB TOKEN HERE"
-
 ```
 
 Set `showGithubProfile` to true or false to show Contact Profile using Github, defaults to false.
 
 Note: Open Source Projects section only show pinned itens of your Github. If you are seeing [this](https://i.imgur.com/Hj6mu1K.png), please follow this
 [Instructions](https://docs.github.com/en/enterprise/2.13/user/articles/pinning-items-to-your-profile).
 
+If the above method by any chance fail, you can visit the [wiki page](https://github.com/saadpasta/developerFolio/wiki/Github-Setup-For-Open-Source-Projects).
```
kartikcho, line 107: Annoying nit: change `fail` to `fails` and remove "you can"

PR #235, opened 2020-08-19
Updates Achievements and Certifications Card UI

zekinah:
Request from Issue #4.

Preview:
image

### DIFF
--- a/src/components/achievementCard/AchivementCard.js
+++ b/src/components/achievementCard/AchivementCard.js
@@ -1,4 +1,5 @@
 import React from "react";
+import "./AchievementCard.css";
 
 export default function AchivementCard({ cardInfo }) {
   function openUrlInNewTab(url) {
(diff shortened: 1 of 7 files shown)


src/components/achievementCard/AchivementCard.js
```diff
@@ -18,7 +19,7 @@ export default function AchivementCard({ cardInfo }) {
         </div>
         <div className="certificate-card-footer">
           {cardInfo.footer.map((v, i) => {
-            return <p onClick={() => openUrlInNewTab(v.url)}>{v.name}</p>;
+            return <span class="certificate-tag" onClick={() => openUrlInNewTab(v.url)}>{v.name}</span>;
```
saadpasta, line 22: Try to add a hover effect

zekinah: Added Hover effect: 2c6b27b

Also did some CSS tweaks on the other buttons.

**Hero div**
image

**Talk div**
image

naveen521kk: The colours in talk div doesn't look good.

zekinah: What if we make it like this @naveen521kk ? the goal of this is to have consistency in the color scheme of the site.

image

And the hover effect of the button is like this.
https://www.loom.com/share/fe444f74e739463fa9c53afd6d7c91d6

saadpasta: @zekinah let's fix this issue in this PR. I guess it's just a background color so fix this in this PR.

zekinah: @saadpasta , ok,  here it is 5953a31.

PR #238, opened 2020-08-30
add warning when podcast does not load

waqasnasir:
add warning message when iframe does not load.

### DIFF
--- a/src/containers/podcast/Podcast.js
+++ b/src/containers/podcast/Podcast.js
@@ -3,10 +3,21 @@ import "./Podcast.css";
 import { podcastSection } from "../../portfolio";
 import { Fade } from "react-reveal";
 
+const cantDisplayError = "<div class='centerContent'><h2>Can't load? Check privacy protection settings</h2></div>";
+
+function timeOut(){
+  setTimeout(function() {
+    if (!(document.getElementById('podcasts').innerHTML.includes('iframe'))){
+      document.getElementById(`podcasts`).innerHTML = cantDisplayError ;
+    }
+  },
+  10000);
+}
+
 export default function Podcast() {
   return (
     <Fade bottom duration={1000} distance="20px">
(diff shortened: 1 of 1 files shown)


kartikcho:
Hi @waqasnasir! It looks like your implementation of the twitter component's warning message doesn't seem to work. You should give it another shot and let's get this merged soon :)

src/containers/podcast/Podcast.js
```diff
@@ -3,10 +3,21 @@ import "./Podcast.css";
 import { podcastSection } from "../../portfolio";
 import { Fade } from "react-reveal";
 
+const cantDisplayError = "<div class='centerContent'><h2>Can't load? Check privacy protection settings</h2></div>";
+
+function timeOut(){
```
kartikcho, line 8: The Twitter message implementation uses [Suspense](https://reactjs.org/docs/concurrent-mode-suspense.html) so you should try to do something similar here too with a fallback loader.

kartikcho (changes requested):
@saadpasta what's up with the Vercel deployments asking for permissions?

src/containers/podcast/Podcast.js
naveen521kk: Actually speaking it uses lazy loading, https://reactjs.org/docs/react-api.html#reactlazy, which in turn uses suspense.  See #181

naveen521kk:
Any update on this @waqasnasir

kartikcho:
@naveen521kk what do you think about closing this to allow new contributors to work on this issue? There seems to be no activity for almost a month on this PR.

naveen521kk:
> what do you think about closing this to allow new contributors to work on this issue? There seems to be no activity for almost a month on this PR.

Ok @kartikcho. If @waqasnasir wants to works on this again he can make a new PR.

PR #254, opened 2020-10-03
simplify code contributions by fully automating the dev setup with Gitpod

nisarhassan12:
This Pr simplifies code contributions by fully automating the dev setup with Gitpod(an online open-source VS Code like ide). With a single click, it will launch a workspace and automatically:

- clone the developerFolio repo.
- install the dependencies.
- start the dev server.

So that anyone interested in contributing or playing around with the code can start straight away without any friction.

You can give it a try on my fork of the repo via the link: 

https://gitpod.io/#https://github.com/nisarhassan12/developerFolio

This is how it looks:

image

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,12 @@
+tasks:
+  - name: Dev Server
+    init: npm install
+    command: npm start
+
+ports:
+  - port: 3000
+    onOpen: open-preview
+
+vscode:
+  extensions:
+    - esbenp.prettier-vscode@5.7.1:4Zx39KyQMoIz7x94PSmDmQ==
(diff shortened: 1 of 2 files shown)


kartikcho:
While I really like the benefits of Gitpod, I think it still works without a config if you just prefix the GitHub URL with `gitpod.io/#`.

I'm not exactly for the decision of adding this but it isn't hurting the project in any way so I don't have a strong opinion.

cc: @saadpasta @naveen521kk

.gitpod.yml
```diff
@@ -0,0 +1,12 @@
+tasks:
+  - name: Dev Server
+    init: yarn install
+    command: yarn start
```
naveen521kk, line 4: Using `npm` is a lot better. AFAIK. because `yarn.lock` is removed in #255

saadpasta:
@nisarhassan12 That's a great idea and Thanks for checking out DeveloperFolio.I think we don't need this for now.

nisarhassan12:
Ok. Thanks.

PR #259, opened 2020-10-08
feat: add prettier and a check for it using github actions

naveen521kk:
Forked from #107 
I added prettier as discussed in #107 and formatted the whole codebase. Also, added a check for pull requests also.

PR #138, opened 2020-05-21
Correction warning and add display option

Exspiravit:
Hi bro,
I add the display option to make them appear and disappear from both the menu and the front page.
Fixed some warnings for using class instead of class name.
I took the liberty of adding the codersrank library so that you can easily add your profile to the portfolio.

PR #260, opened 2020-10-11
Add lottie illustrations

### LINKED ISSUE #257: Add Lottie illustrations behind a preference
The current illustrations, although look good with the current aesthetic, are static and don't sit right with the recent 'interactive' animations added. 
We can add a switch to use [Lottie illustrations](https://github.com/airbnb/lottie-web) instead of the current ones based on what the user decides in `portfolio.js`.
This does have a downfall, Lottie is pretty performant but still might have performance issues running on slower systems. Hence, the switch to use it or not.

cc: @saadpasta @naveen521kk

kartikcho:
Resolves #257 
Following feedback and discussion in the issue, I've added Lotties instead of our current illustrations. I've also added a switch in `portfolio.js`, if the user chooses to use images instead. 

~~There's a slight issue with the `react-lottie` library, on clicking the animation it pauses it (super weird right?). I looked a lot into the docs but couldn't find anything to go around this, so I implemented my own workaround for the clicks. The problem is, it tends to 'hiccup' the animation for 0.5 second or so, which isn't a big deal but should be known. I'm open to better suggestions!~~

Turns out I was only experiencing it in dev and it works perfect in the prod preview!


naveen521kk (approved):
Looks Nice. It should wait for #92 and #259 
Much needed changes in `src/portfolio.js`.

kartikcho:
Thanks. We should get this in before #259 though since it'll reformat the whole code base at once.

WarenGonzaga:
This is cool @kartikcho

WarenGonzaga:
I guess you need to fix some conflicts on your PR @kartikcho

kartikcho:
I know 😅, like I said before there's a sequence of PRs being blocked by one PR, i.e #138. 
Waiting for that blocker to be resolved soon.

WarenGonzaga:
Agree @kartikcho

saadpasta:
@kartikcho please fix the conflicts so I can review this. 
Thanks 👍

kartikcho:
There's a strange rendering bug with the lottie player after integrating the recent changes. Will take a deeper look by this weekend.

saadpasta:
@kartikcho it's working on dev?

kartikcho:
The bug is visible on vercel's deployment of this branch. 3 of them don't play on load (strangely the last one does) and the on click pause is persistent again.

Also, will have to reverse look up the Lottie I picked, because 2 of them have white backgrounds visible in dark mode.

kartikcho:
@saadpasta @naveen521kk updated the PR with no new bugs (to the best of my knowledge lol), have a look!

naveen521kk (changes requested):
**SUGGESTION**: Create a `.gitattributes` file with values `src/assests/lottie`, assuming it is autogenerated, this prevents it from appearing it on code reviews, see https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/customizing-how-changed-files-appear-on-github

naveen521kk (approved):
Other than previous comment, LGTM!

saadpasta (approved):
LGTM ✌. 
Just fix the review from @naveen521kk .

kartikcho:
@naveen521kk the lottie isn't auto generated, it's basically JSON data needed to animate the illustration, which is required. 

About appearing in diffs, I agree but what if a lottie is updated and we miss it in a PR? Its kind of a rare edge case scenario but is still a possibility, besides we won't need to update it again for a while.

naveen521kk:
If it is autogenerated one then no need to do so. So, LGTM, merging.

Issue #263, opened 2020-10-14
Dark mode slider looking weird


fabianchua6:
**Describe the bug**
Dark mode slider is fine when it's in dark mode but when toggled to light mode, the toggle leaves a weird gap in the slider itself.

**To Reproduce**
1. Click on dark mode toggle to light mode.
2. See error where there is a gap between the toggle and the slider

**Screenshots**
![image](https://user-images.githubusercontent.com/57614749/96021085-8a5f3f00-0e81-11eb-94b8-db999b44d7a8.png)
As you can see, there is abit of blue on the right of the toggle.

kartikcho:
Ah, I noticed this too but it might've slipped through. Would you like to open a PR for this @fabianchua6 ?

WarenGonzaga:
@kartikcho can I take this issue?

fabianchua6:
Sure please!

kartikcho:
@fabianchua6 just confirming if you wanted to take the issue or are okay with someone else making a PR for it?

WarenGonzaga:
I'm working on it @kartikcho @fabianchua6

fabianchua6:
> @fabianchua6 just confirming if you wanted to take the issue or are okay with someone else making a PR for it?

I’m fine with it! Thanks for asking 👍🏻

WarenGonzaga:
Hi @saadpasta done with this issue kindly review it I add additional update which is Vulnerability fixes on your npm packages.

WarenGonzaga:
@saadpasta I guess this is done? you can close now.

saadpasta:
@WarenGonzaga We can close the issue only after the PR is merged.

kartikcho:
fixed in #264

PR #264, opened 2020-10-15
Dark Mode Slider Fix with Security Fixes

WarenGonzaga:
This is my commit to fix the security vulnerabilities.

PR #265, opened 2020-10-16
Fix dark mode label

sarveshh:
Solved issue #263 "Dark mode slider looking weird".
Dark mode slider was off by 2px when slided to the right.
Please review.


kartikcho:
Hey @sarveshh, we'll have to close this PR down sadly as #264 has fixed the same and the issue was assigned to it's author. You can look into other issues though!

PR #266, opened 2020-10-17
Fix/component name

Rizwanjamal:
**What changes this PR contains:**

That PR contains the fix for the spelling mistakes in word _**Achievement**_ that's being used in codebase on different places


kartikcho (approved):
Ah, good catch. Thanks for the fix!

PR #274, opened 2020-11-22
add popular social media selection

jahwanoh:
I wanted to add my youtube, and I think there would be needs for others too.

new: youtube, discord, whatsapp, soundcloud, tumblr, snapchat, twitch, slack, vimeo
color: picked by its official logo


saadpasta (commented):
Thank you so much for contributing to DeveloperFolio, but I think we don't need this right now.

PR #281, opened 2020-12-09
Modern website container added

kaustubhai:
Padding from both sides of screen added to make it look more modern

### DIFF
--- a/src/containers/Main.css
+++ b/src/containers/Main.css
@@ -3,3 +3,20 @@
   color: white;
   transition: "0.1s";
 }
+
+.containerContent{
+  padding: 0 10%;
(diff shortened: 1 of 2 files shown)


saadpasta (commented):
Thanks for contributing but I don't think we could merge this.
Thanks

kaustubhai:
I'm sorry but since it was my first PR here, it will be of a lot of help if you can tell me what went wrong so that  I can take care of it next time

saadpasta:
@kaustubhai Before creating a PR you should create an Issue and then create a PR or work on that PR that has an issue that would help you get your PR accepted more often

PR #311, opened 2021-02-01
feat: storing theme in local storage

yogeshhrathod:
saving the theme user already selected so next time that theme will be appeared

### DIFF
--- a/src/containers/Main.js
+++ b/src/containers/Main.js
@@ -27,7 +27,7 @@ export default class Main extends Component {
   }
 
   componentDidMount() {
-    if (localStorage.getItem('isDark')==null) {
+    if (localStorage.getItem('isDark')===null) {
       const darkPref = window.matchMedia("(prefers-color-scheme: dark)");
       localStorage.setItem('isDark',darkPref.matches)
     }  
--- a/src/containers/Main.js
+++ b/src/containers/Main.js
@@ -17,7 +17,6 @@ import Twitter from "./twitter-embed/twitter";
 import { StyleProvider } from "../contexts/StyleContext";
 import "./Main.css";
 import Profile from "./profile/Profile";
-import { educationInfo } from "../portfolio";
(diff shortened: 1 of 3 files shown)


src/containers/Main.js
```diff
@@ -28,11 +27,17 @@ export default class Main extends Component {
   }
 
   componentDidMount() {
-    const darkPref = window.matchMedia("(prefers-color-scheme: dark)");
-    this.setState({ isDark: darkPref.matches });
+    if (localStorage.getItem('isDark')==null) {
```
saadpasta, line 30: Please use strict equality comparison (===) with indentation.

yogeshhrathod: I have fixed this ✔
Thanks

saadpasta (commented):
Please remove `README.md` file.

kartikcho:
@saadpasta do we need this? There was another PR I think that tried to add this feature but we rejected it. 
I think that no one will come to the same portfolio more than once in a long time, so leaving cache just to save a button click to set the theme in someone's browser is a very rare use case.
I don't want the work that went into this PR to go to waste but then this wasn't a planned issue afaik. Open to hearing your thoughts but I don't think we need this.

yogeshhrathod:
You are right @kartikcho, but if somebody checking out the portfolio she/she will not know the you motto of saving cache. and without this feature its looks buggy, people will not revisit the page but they can refresh that page and there the can think that is as misbehavior
Thanks

kartikcho:
That is a fair argument @yogeshhrathod and could be useful. But personally I don't think this is necessary.

This isn't a big change so I'm not feeling strongly against this, I guess you can take the final decision @saadpasta

saadpasta:
@kartikcho I think it's a small feature and won't impact much and it's good to have.

saadpasta (approved):

PR #318, opened 2021-02-04
Added section for Codersrank widgets

akashp1712:
- Added Github Activity interactive timeline widget from Codersrank


akashp1712:
I'm not a pro Reatjs developer. I have a beginner level experience.
I've included the Codersrank widget in my website, hence raise a PR back to the main repo.

kartikcho:
Hey @akashp1712 , thanks for opening a PR for this!

It's alright if you're not well experienced with React, it's great that you worked on adding this. Although, in the future opening an issue with a suggestion to add something would be better so we'll know what we're expecting and your hardwork isn't wasted if we don't want to add it.
 
Can you send some screenshots before @saadpasta authorizes the preview.

akashp1712:
Thanks, @kartikcho for the review. Sure, I will inform in advance if I'm working on a feature.
Please find the screen below for the current implementation:

codersrank-github-activity

Please review my website: https://akashpanchal.com for the same.

public/index.html
```diff
@@ -38,6 +38,7 @@
 
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/FortAwesome/Font-Awesome@5.13.0/css/all.min.css">
 
+    <script src="https://unpkg.com/@codersrank/activity@0.9.14/codersrank-activity.min.js"></script>
```
kartikcho, line 41: We should download this file, move it into a `utils` folder and import it from there so it can get optimized during compile time.

kartikcho (changes requested):
This looks neat and could be a great addition! 
Without looking much under the hood and assuming from your fork of the project, this needs to accommodate for the following cases:

- Dark mode of the app isn't sitting right with the component, might have to tweak a bit of CSS to take that into account.
- On smaller screens, less than 768px, hide the component because it is barely visible to make sense of it. Would be open to thoughts from @saadpasta @naveen521kk 
- Move this component with the github projects section where it will make much sense (try using same keys for stuff in `portfolio.js` too wherever possible).

Also, can we remove that 'codersrank' watermark? I don't think people would like that a lot, if not then we can find another open source alternative that doesnt have a watermark.

kartikcho:
I think I asked a lot from you, let me know if you're stuck somewhere. Let's get this PR to the finish line!

public/index.html
naveen521kk: well it's in npm https://www.npmjs.com/package/@codersrank/activity I think that should be used.

src/portfolio.js
```diff
@@ -238,6 +238,12 @@ const bigProjects = {
   display: true, // Set false to hide this section, defaults to true
 };
 
+const codersRank = {
+  userName: 'YOUR_USERNAME', // Change to your codersRank username to display widgets from codersRank
+  githubActivityChart: true, // Set true or false to show Github Activity Chart
+  display: true, // Set false to hide this section, defaults to true
```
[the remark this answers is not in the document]
akashp1712: yes, I've used two flags considering extending the section with more widgets form Codersrank in future: [](https://docs.codersrank.io/widgets/)

src/containers/codersrank/Codersrank.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import './Codersrank.css';
+import { codersRank } from '../../portfolio';
+
+export default function Codersrank() {
```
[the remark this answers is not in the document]
akashp1712: Yeah, Could you please take a look at the other widgets as well here: [](https://docs.codersrank.io/widgets/)
If we don't want other than the Github Activity chart then we can proceed with your suggestions.

public/index.html
akashp1712: Yes, It is in npm. I'll update this.

akashp1712:
Followings are the checklist of the review comments:

- [x] 1. remove watermark
- [x] 2. use NPM library
- [ ] 3. Dark mode support - I would require some guidance here
- [ ] 4. smaller screen support - I would require some guidance here
- [ ] 5. Move this to Github project section - Please advice on this whether to keep it separate for more widgets in future or not

kartikcho:
Your progress so far has been great!

> Dark mode support

For this, look at how existing components are rendering when dark mode is toggled. We're getting the mode toggle state through the context API and giving appropriate classes to the component accordingly. 
Look at Header here for example: 
https://github.com/saadpasta/developerFolio/blob/master/src/components/header/Header.js#L27
You'll need to write some CSS for how the component would look like in dark mode.

> smaller screen support

This can be handled with a simple CSS media query to add `display: none;` on screens smaller than 768px (or whatever the exact pixel width is of the phone screen breakpoints)

> Move this to Github project section
Sorry for being unclear earlier, what I meant by this was to move it below the Github projects component in terms of order of components being rendered. This component should be separate from the other components though.

Let me know if you need more help!

akashp1712:
Following comments were addressed:

- [x] 1. remove watermark
- [x] 2. use NPM library
- [ ] 3. Dark mode support - I tried but couldn't override the default css. We may have to assign it to someone with better CSS grasp
- [x] 4. smaller screen support
- [x] 5. Move this to Github project section
- [x] 6. Rename the section to GithubActivity

src/portfolio.js
```diff
@@ -238,6 +238,11 @@ const bigProjects = {
   display: true, // Set false to hide this section, defaults to true
 };
 
+const githubActivity = {
+  userName: 'YOUR_USERNAME', // Change to your codersRank username to display widgets from codersRank
```
saadpasta, line 242: @akashp1712 change it saadpasta so we can view the activity. 
Q: Do I need to have codersRank profile to view the Github Activity

akashp1712: Yes @saadpasta, you will require to have the **codersrank** account. As they are rendering the GitubActivity component.

kartikcho: If it won't work without a profile on that application then it isn't really user friendly imo. Since we're already asking for Github token from users, we can use an open source GitHub graph component (if there is any) to save the trouble of making an account for this feature. Let me know what you think.

saadpasta: @akashp1712 would agree with @kartikcho could you look up some other library for this.

akashp1712: Yeah, agree with this as not all user would go and make an account in the third-party apps.
This was the only available solution with interactive github activity widget (afaik). We can close this pull until we find something similar.

kartikcho: I'd really hate for your work to go to waste but this could really be a great addition next to our GitHub projects section. If you could find any alternatives that work using gh access tokens, definitely open a PR for adding it (or an issue first to make sure it doesn't go to waste).

I'll close it for now but feel free to look into other issues if you want to dabble more with React!

PR #326, opened 2021-02-10
Run npm audit fix, manually fix some warnings

kartikcho:
Fixed some package vulnerabilities and React warnings. One package, `react-scripts`, has a major version update that has breaking changes so I have knowingly left it out. 
Updating that package without knowing the full extent of changes would be a bad decision (and one I'm avoiding right now).

### DIFF
--- a/src/components/header/Header.js
+++ b/src/components/header/Header.js
@@ -25,7 +25,7 @@ function Header() {
   return (
     <Headroom>
       <header className={isDark ? "dark-menu header" : "header"}>
-        <a href="" className="logo">
+        <a href="/" className="logo">
           <span className="grey-color"> &lt;</span>
           <span className="logo-name">{greeting.username}</span>
           <span className="grey-color">/&gt;</span>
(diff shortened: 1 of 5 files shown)


src/components/header/Header.js
```diff
@@ -25,7 +25,7 @@ function Header() {
   return (
     <Headroom>
       <header className={isDark ? "dark-menu header" : "header"}>
-        <a href="" className="logo">
+        <a href="/" className="logo">
```
naveen521kk, line 28: this should be better.
```suggestion
        <a href="#" className="logo">
```

naveen521kk (changes requested):

src/components/header/Header.js
kartikcho: I actually was going to go with "#" but I think in this case "/"  would be better as it will refresh the page, which is the point of clicking on a "home" button. 

Let me know what you think about this.

naveen521kk: Ok, this looks good then. :)

naveen521kk (approved):

Issue #320, opened 2021-02-05
Token revoked when deploying to gh-pages 


martiriera:
Hello Saad and the rest of contributors. First of all congratulations for this work, very fancy stuff!

I have a master branch with a source code and a gh-pages one for the deployment. The thing is that when I execute the `npm run deploy` from master branch, GitHub detects the access token on a commit and revokes the token.

As stated on issue #180 I have the .env file on the .gitignore of the master branch. I am also defining the secret on the repo.

Any idea? I am missing something?

Thank you.

Matesanz:
It is also happening to me. If I use the **token while developing there's no problem**, everything works fine.
If I run the `npm run deploy` and before entering the credentials I run `serve -s build` everything is ok.
BUT, **when deploying to gh-pages** the problem faces: the **token disappears...**

saadpasta:
@martiriera make sure your .env file is in .gitignore. Try to generate a new token and try again or use another method of converting the token to base64 and then converting it back to normal when using so Github doesn't detect it

martiriera:
@saadpasta thanks for your answer. 

I am sure that my .env is not pushed. What is really happening is that the token appears in plain text on a file commited in "gh-pages", here an example: [https://github.com/martiriera/martiriera.github.io/blob/8a7caed452fca1217a9c11f1eed8c772eb196cd9/static/js/main.be129e4c.chunk.js ](url)
I tried to encrypt the token when using it on the app but it continues appearing on this "main" chunk. 

Maybe my deployment is not correct? It's something related to the repo secret?

Thanks for your time.

harikesh409:
As stated in the [create-react-app](https://create-react-app.dev/docs/adding-custom-environment-variables/) the token will be visible in the static files once it is builded, that is the reason github is removing it.

```
WARNING: Do not store any secrets (such as private API keys) in your React app!
Environment variables are embedded into the build, meaning anyone can view them by inspecting your app's files.
```

martiriera:
@harikesh409 thanks for your answer.

As stated [here](https://stackoverflow.com/a/57103663/14983877) the only way to do so is making a request to a back-end API. Is there a way to do it with the tools used on this portfolio (GitHub Actions, repo secrets)?

naveen521kk:
> Is there a way to do it with the tools used on this portfolio (GitHub Actions, repo secrets)?

[...]

PR #333, opened 2021-02-27
feat: store profile information locally

naveen521kk:
This is better than requesting information using GitHub client side
Fixes https://github.com/saadpasta/developerFolio/issues/320
Also, fixes various warning during build

Something to note here, I wasn't able to use `.env` files, I think that requires configuration using https://github.com/motdotla/dotenv
but for now, works when environment variables are set manually.


naveen521kk:
@saadpasta you would need to set an env var called `GITHUB_USERNAME=saadpasta` in vercel's configuration.

src/containers/profile/Profile.js
```diff
@@ -13,46 +12,29 @@ export default function Profile() {
...
+            }
+            console.error(result);
+          })
+          .then(response => {
+            setProfileFunction(response.data.user);
+          })
+          .catch(function (error) {
+            setProfileFunction("Error");
+            console.log(
+              "Because of this Error Contact Section is Showed instead of Profile"
```
kartikcho, line 32: nit: We can maybe change this to something like 'Because of this error, contact section has reverted to default' or something since both the variants are for contact section.

kartikcho (approved):
Everything looks good to me and _should_ work. I would've held back on getting this in until I saw a preview deployment @saadpasta but I guess it won't be visible until the new env variables are setup.

src/containers/profile/Profile.js
naveen521kk: yeah, I didn't change these actually, I just copy-pasted how it was previously, let me change it.

naveen521kk:
I think it should work, I tried to deploy it in another branch using vercel. https://github.com/naveen521kk/developerFolio/commit/199d0ee6e60e87b4d4bdfac6d79ee5e068180d12#commitcomment-47636294

kartikcho:
Your deployed commit looks good, but will hold until Saad configures Vercel accordingly for these changes. Also, we should update docs to reflect this new deployment flow if there are some misleading instructions from the old flow @naveen521kk.

saadpasta (commented):
@naveen521kk this looks good but don't you think that hitting deployment after every change in your pinned repo and profile is kind of tedious. Is there any other workout for this because I don't think this is a good approach this will cause confusion for new people using this project and will also limit us when we try to use any other Github API in the future for this project. 

We could use Github converted token approach that we were using in the past to make sure the token doesn't get revoked I am using a similar approach in another open-source project [react-blog-github](https://github.com/saadpasta/react-blog-github/blob/master/src/Utils/apollo.js) and it's working fine. 

Tell me what you guys think 🙂

naveen521kk:
Hi Saad,
That was previously discussed over https://github.com/saadpasta/developerFolio/issues/320

> don't you think that hitting deployment after every change in your pinned repo and profile is kind of tedious. 

IMHO, that should be the best way to go about it. I don't think people usually change pinned repositories or profile more often, even if they did so, it should fine for them to click on a button in the repo and GitHub Actions refreshing it should be doable. See https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch
Or rather if it is someone who changes them frequently then they could set up a CRON job on Github Actions, which will redeploy the site, whenever they wanted to. https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events

> Is there any other workout for this because I don't think this is a good approach this will cause confusion for new people using this project

I don't think there should be any confusion that would happen. Essentially, people wouldn't know about this change unless they are changing their profile to often or something, in that case, they could look over the docs for how to. (yeah, we should document that.)

> will also limit us when we try to use any other Github API in the future for this project.

That shouldn't be the case. We could edit `fetch.js` to get the additional information and save it in the same file or rather create a new file like that, that shouldn't matter.

> We could use Github converted token approach that we were using in the past to make sure the token doesn't get revoked I am using a similar approach in another open-source project react-blog-github and it's working fine.

In general, storing Github's Personal Access tokens publicly isn't that great, even if it doesn't have any permissions. Anyone could get that token easily by inspecting the request, that's the problem, and the token can be misused. I would simply say, the things I did in this PR is the best for static sites without any backends.

saadpasta:
@naveen521kk If the token is without any permission I don't know how it could be misused. If anyone would get the token he would only view public info nothing more than that.

naveen521kk:
Quick question, why does Github automatically remove it in that case? There could be something that could be dangerous, and we don't want that to happen to users.

naveen521kk:
Also, @saadpasta I think you didn't set up an env var `GITHUB_USERNAME=saadpasta`

saadpasta:
@naveen521kk Yeah I guess you are right but I am thinking there must be another way to do right ??? Like how do we secure the token.. Actually, I am not up for prefetching this for now. Let's see If we can find something otherwise I think we need to go with this

kartikcho:
> If anyone would get the token he would only view public info nothing more than that.

It's still not a good thing to make the token public from a security perspective and also development standards. 

Honestly I don't have a problem with this approach as long as it is well documented for the users to figure out on their own. 
Can we default to a weekly scheduled action trigger in the config to make setup a bit easier for every user?

naveen521kk:
Can we leave the documentation for another PR?

.gitignore
```diff
@@ -21,3 +21,6 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+#user genereated content
```
kartikcho, line 25: nit: spelling mistake in 'generated'

kartikcho (commented):
We can skip documentation for now but we should follow up with a PR to cover it soon. 
Also, I don't see code for a scheduled workflow in these changes, can we add a weekly default job too?

naveen521kk:
>  Also, I don't see code for a scheduled workflow in these changes, can we add a weekly default job too?

will do soon.

saadpasta:
> Also, @saadpasta I think you didn't set up an env var `GITHUB_USERNAME=saadpasta`

Can I setup an env for this PR?

naveen521kk:
> Can I setup an env for this PR?

That should be fine. If everything looks good can this be merged @saadpasta?

saadpasta (approved):

saadpasta:
Finger Crossed 🤞 . Let's deploy 😃

naveen521kk:
Looks like it has some problem with env vars, I think I know to fix them, let me make a quick PR.

PR #334, opened 2021-03-07
fix(node.js.yml): set GITHUB_USERNAME env var

naveen521kk:
see https://github.com/saadpasta/developerFolio/pull/333#issuecomment-792270471


saadpasta (approved):

naveen521kk:
I still wonder what happens and why Github returning 401. Maybe the token you set in the secret panel expired or something @saadpasta ?

saadpasta:
I will try to change the token and redeploy

PR #336, opened 2021-03-07
fix(fetch.js): load .env variables

elvisciuffetelli:
Hi, I had issues loading the .env variables which resolved in undefined, I resolved with this

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     ]
   },
   "devDependencies": {
-    "dotenv-webpack": "^7.0.1",
+    "dotenv": "^8.2.0",
     "prettier": "^2.2.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     ]
   },
   "devDependencies": {
+    "dotenv-webpack": "^7.0.1",
     "prettier": "^2.2.1"
   }
 }
--- a/fetch.js
+++ b/fetch.js
@@ -1,6 +1,7 @@
 fs = require("fs");
 const https = require("https");
 process = require("process");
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -40,6 +40,7 @@
     ]
   },
   "devDependencies": {
+    "dotenv-webpack": "^7.0.1",
```
naveen521kk, line 43: why not use the latest version? `8.2.0`

naveen521kk (changes requested):

package.json
elvisciuffetelli: I think this is the latest version 
image

naveen521kk: why `dotenv-webpack` instead of `dotenv`?
https://www.npmjs.com/package/dotenv

elvisciuffetelli: `dotenv` is used if you want to use variables from your .env file in the webpack.config.js directly and it behaves just as any other javascript file while being run. `dotenv-webpack` plugin instead inserts the variables into your bundled code, which is enough for the deploy

naveen521kk: If you looked into the file you are changing `fetch.js` it is doing it for once, once it fetches the details those env vars are used at all. If it runs before starting the react-scripts build. Essentially, just because there was a vulnerability in inserting the env vars in build code I hade to create #333 so it should be fine to use `dotenv` here.

elvisciuffetelli: ok, changed it :)

naveen521kk (approved):

naveen521kk:
@all-contributors please add @elvisciuffetelli  for code

Issue #322, opened 2021-02-08
Dynamic Color Scheme 


saadpasta:
## Summary
Currently, we are only one single color scheme in DeveloperFolio. We need to make it dynamic. 

### What's new?
We need to create a global CSS variable or something like that portfolio,js so we can get the same color across all components 

### Motivation
So a user could change the color scheme to their own preference

kartikcho:
Assigning this to @SJellen

kartikcho:
This has been delivered in #355

PR #352, opened 2021-04-11
Dynamic color scheme

SJellen:
Dynamic Color Scheme #322

I've moved all the color styles to one global file so that a user can change their colors from that file. But in order to do that I need to add sass and change the Css to Scss. 

It's really the first step for someone to make it more dynamic but at least now you change change all instances of a color at once.


kartikcho:
Hey @SJellen, sorry for the late reply and thanks for working on this issue, it'll be a big addition to the project!

Took a brief look at the changes, I see some keys with same values like our portfolio's `#55198b` color. We could either unify it into one single theme color 'switch' or keep them individually as is for more control but more hassle for some users.
If possible (I don't have much experience building with Sass), we could add a `globalThemeColor` key which, if configured, would override all the other individual color values. This would be a best of both worlds solution for this if possible, let me know if you have any questions or suggestions.

Also, @saadpasta can you authorize the preview deployment so we can review and get this in soon!

kartikcho:
Hey @SJellen any reason why you closed your WIP PR?

naveen521kk:
Looks like it is moved to another PR #355

kartikcho:
Ah nice, didn't notice

PR #355, opened 2021-04-20
Dynamic color scheme #352

SJellen:
I've cut down the duplicates. 
So basically a user can change their colors from the _globalColor file. 
But here is something to think about... 

1. The total colors that are used right now are around 50.  I think you could cut about 20 extra colors and not lose anything. But it's not my repo and I wouldn't feel comfortable doing that.
2. When a user changed their color scheme it won't match the highlight colors of the stock images.


package.json
```diff
@@ -13,7 +13,8 @@
     "react-lottie": "^1.2.3",
     "react-reveal": "^1.2.2",
     "react-scripts": "^3.4.3",
-    "react-twitter-embed": "3.0.3"
+    "react-twitter-embed": "3.0.3",
+    "sass": "^1.32.8"
```
naveen521kk, line 17: why not use `node-sass`? any problems?

kartikcho:
Really great addition, appreciate the work!

> The total colors that are used right now are around 50. I think you could cut about 20 extra colors and not lose anything.

If we could group them together like all button colors under a single 'key', it'd be much better and easier to customize. 
Maybe leave a comment on how to add and use custom variables as this project can still sometimes be inaccessible to some new users.

> When a user changed their color scheme it won't match the highlight colors of the stock images.

This is something we can't solve programmatically, at least not for the static assets used. Even for Lottie it'll be more hassle than worth for now but maybe we can look into it in the future. Generally we still recommend users to use their own assets for creating a more unique portfolio.

kartikcho:
Hey @SJellen! Checking in to see if you're doing alright and still working on this PR.

SJellen:
Sorry I've just been really busy. 
I think I've found a way though to drop the Sass and revert to css but keep the variables. So thats what I'm tinkering with.

src/_globalColor.scss
```diff
@@ -0,0 +1,87 @@
...
+$github: #333;
+$gitlab: #fca326;
+$google: #ea4335;
+$twitter: #1da1f2;
+$medium: #000;
+$stackoverflow: #f48024;
+$instagram: #c13584;
+
+
+
```
naveen521kk, line 87: can you remove extra lines here?

src/_globalColor.scss
```diff
@@ -0,0 +1,87 @@
...
+$lightBackground3: #f5f2f4;
+$blogCardContainerColor: #586069;
+// dark background colors 
+$darkBackground: #171c28;
+
+// light theme box shadows
+$lightBoxShadow1: rgba(0, 0, 0, 0.2);
+$lightBoxShadow2: rgba(0, 0, 0, 0.1);
+$lightBoxShadow3: rgba(0, 0, 0, 0.3);
+$lightBoxShadow4: rgba(0, 0, 0, 0.08);
```
naveen521kk, line 45: It would nice to add a comment about where these colours are used  are used.

naveen521kk (approved):
LGTM

kartikcho:
Don't merge this yet, I'd like to review it too before we get this in. 
Also, fix the formatting test check by running format on your fork.

src/_globalColor.scss
kartikcho: We can rename these to:
 - `$lightBoxShadow1` -> `$lightBoxShadowDark` 
 - `$lightBoxShadow2` -> `$lightBoxShadow` 
 - `$lightBoxShadow3` -> `$lightBoxShadowDarker`
 - Remove `$lightBoxShadow4` because it is only being used in `Blog.scss` and can be replaced with current `$lightBoxShadow2` as the colors are very close shades of each other.
 
The advantage of renaming it this way would help people who want to use these variables in the future to design their own components without getting confused.

src/_globalColor.scss
```diff
@@ -0,0 +1,84 @@
...
+// text colors light theme 
+$titleColor:  #000000;
+$textColor: black;
+$subTitle: #868e96;
+$cardSubtitle: #666666;
+$talkCardSubTitle: #7f8287;
+$blogCardTitleColor: #262626;
+
+// text color dark theme 
+$textColorDark: white;
```
kartikcho, line 19: Nit/Suggestion: We can move to only using hex codes for color variables.

src/containers/blogs/Blog.scss
```diff
@@ -1,18 +1,20 @@
...
+  box-shadow: 0px 20px 50px v.$darkBoxShadow;
   -webkit-transition: all 0.3s ease;
   -o-transition: all 0.3s ease;
   transition: all 0.3s ease;
   margin-top: 30px;
 }
 
 .square:hover {
-  box-shadow: 0 20px 40px rgba(0, 0, 0, 0.08);
+  box-shadow: 0 20px 40px v.$lightBoxShadow4;
```
kartikcho, line 17: You should change this to `$lightBoxShadow2` and remove `$lightBoxShadow4` variable altogether as I mentioned in one the previous comments.

src/_globalColor.scss
```diff
@@ -0,0 +1,84 @@
...
+$educationCardBorder: #a9a7f9;
+
+// hovers
+$achievementCertificateCardHoverDark: rgba(255, 255, 255, 0.2);
+$blogCardHoverSmallColor: rgba(255, 255, 255, 0.8);
+$headerHoverBG: #f4f4f4;
+$contactDetailHoverTS: #b5b5b5;
+
+// misc backgrounds
+$meterSpanBG: #aaa5ff;
```
kartikcho, line 67: We can change this variable name to `$progressBarSpanBG` and similar for the one on line 56 to `$progressBarBackground`

kartikcho (changes requested):
Sorry for the late review @SJellen, I've been a bit busy for the past few weeks. The PR is amazing and will be a big milestone feature for this project!

Just need to address the review comments along with these two remaining tasks and we should be good to go!

- Run `npm run format` to fix the failing formatting tests.
- We should add a bit of documentation section or description about setting custom colors in `README.md` so new users will know where to tweak these variables from and how.

package.json
```diff
@@ -16,8 +16,9 @@
     "react-lottie": "^1.2.3",
     "react-reveal": "^1.2.2",
     "react-scripts": "^3.4.3",
-    "react-test-renderer": "^17.0.2",
-    "react-twitter-embed": "3.0.3"
+    "react-twitter-embed": "3.0.3",
+    "sass": "^1.32.8",
+    "react-test-renderer": "^17.0.2"
```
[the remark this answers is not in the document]
SJellen: I'm just going through these one by one so you may see a bunch of  commits.
This section is set. 
The react-twiiter is a  result of a `npm audit fix` so if you can drop it on your end go ahead. I cant change it on my end.

I added pretty much the same directions to the readme as I did to the portfolio.js . If you have a better sentence let me know and I'll add it.

Screen Shot 2021-05-16 at 11 46 11 AM

src/App.scss
```diff
@@ -1,3 +1,5 @@
+@use './globalColor' as v;
```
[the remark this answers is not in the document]
SJellen: This part is all set on the next commit.

kartikcho:
Prettier really hates this PR lol. 
@naveen521kk do you know why the format test has been failing? Can we also make the workflow print logs while running?
I ran the same command locally and it says everything is good.

kartikcho (approved):
Everything looks great, thanks for this amazing PR!

naveen521kk:
> do you know why the format test has been failing? Can we also make the workflow print logs while running?

Line endings? Make sure you have LF line endings.

kartikcho:
I believe they are LF, can you take a look at what's up @naveen521kk? I'm out of ideas

naveen521kk:
Looks like there is some problem here. I think this PR can be merged now, and I will make another PR to fix this issue. This is obviously because of line-endings.

kartikcho:
@allcontributors please add @SJellen for code and design!

Issue #358, opened 2021-04-24
Open for Opportunities shows "No" even if the value is true 


jelonmusk:
**Describe the bug**
The github profile card field for "Open for Opportunities "shows "No" even if the value is "true"

**Expected behavior**
The field must display YES rather than a No.

**Is this responsiveness Issue**
NO
**Screenshots**
![Screenshot (56)](https://user-images.githubusercontent.com/33628350/115959035-cefe1e00-a527-11eb-9aee-ad4f3254d85f.png)
![Screenshot (57)](https://user-images.githubusercontent.com/33628350/115959038-d0c7e180-a527-11eb-8679-04342dd89a77.png)
![Screenshot (58)](https://user-images.githubusercontent.com/33628350/115959152-5b104580-a528-11eb-89e2-0ea1b6e0b8ef.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser : Chrome
 - Version : Version 90.0.4430.85

naveen521kk:
Looks like it was broken in https://github.com/saadpasta/developerFolio/pull/333
Found a fix.

naveen521kk:
@jelonmusk can you check if #359 fixes your issue?

jelonmusk:
> @jelonmusk can you check if #359 fixes your issue?

Thanks! 
This works perfectly fine.

jelonmusk:
Hi,
It's been a while since I have used Github pages.
I tried deploying it on "https://username.github.io/"
and turns out the site shows readme content instead of the portfolio.
Any help would be appreciated.

naveen521kk:
@jelonmusk you should either deploy the contents of the build page locally or add with the `.github` folder which has actions which will automatically deploy on push.

jelonmusk:
Thanks! It's all fixed now.

PR #359, opened 2021-04-24
`prof.hireable` isn't defined

naveen521kk:
Check for `prof.isHireable` instead.
Fix https://github.com/saadpasta/developerFolio/issues/358


kartikcho (approved):

src/components/githubProfileCard/GithubProfileCard.js
```diff
@@ -1,12 +1,12 @@
-﻿import React from "react";
+import React from "react";
```
naveen521kk, line 1: this change looks artificial maybe due to line endings.

PR #360, opened 2021-05-01
Fix tests and test errors, move gh fetch request behind env key

### LINKED ISSUE #349: npm run test is failing
**Describe the bug**
when tests are run, select all, the test suite for App.test.js fails

**To Reproduce**
1. Go to command line
2. install all dependencies by running 'npm install'
3. run the following command 'npm run test'
4. select all test cases
5. See error " Cannot set property 'fillStyle' of null"

**Expected behavior**
the test should pass

**Is this responsiveness Issue**
NO

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser CHrome

kartikcho:
This PR resolves #349 by fixing missing test dependencies, jest configs and fixing test errors.

Also moved the GitHub profile fetch script behind an env key which is set to false as default. 
Reason for moving fetch request behind a key is to help setup first dev environment faster and without crashes so new users don't feel discouraged. This would also emphasize properly setting up env variables.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -3,7 +3,7 @@ env:
   CI: false
   GITHUB_USERNAME: ${{ github.repository_owner }}
   REACT_APP_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Use the generated personal access token and add it to repository secrets with the name OPEN_SOURCE_TOKEN
-  USE_GITHUB_DATA: ${{ secrets.USE_GITHUB_DATA }}
+  USE_GITHUB_DATA: "true"
 on:
   push:
     branches:
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -3,6 +3,7 @@ env:
   CI: false
   GITHUB_USERNAME: ${{ github.repository_owner }}
   REACT_APP_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Use the generated personal access token and add it to repository secrets with the name OPEN_SOURCE_TOKEN
+  USE_GITHUB_DATA: ${{ secrets.USE_GITHUB_DATA }}
 on:
   push:
     branches:
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -2,7 +2,7 @@ name: Build and Deploy
 env:
   CI: false
   GITHUB_USERNAME: ${{ github.repository_owner }}
-  REACT_APP_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Use the generated personal access token and add it to repository secrets with the name OPEN_SOURCE_TOKEN
+  REACT_APP_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This is automatically set by Github Actions
   USE_GITHUB_DATA: "true"
 on:
   push:
--- a/README.md
+++ b/README.md
@@ -77,10 +77,17 @@ $ git clone https://github.com/saadpasta/developerFolio.git
 # Go into the repository
 $ cd developerFolio
 
+# Setup environment variables
+
+# For Linux
+$ cp env.example .env
+# For Windows
+$ copy env.example .env
+
 # Install dependencies
 $ npm install
 
-#Start's development server
+# Start a local dev server
 $ npm start
 ```
 
@@ -88,9 +95,9 @@ $ npm start
 
 Generate a Github personal access token following these [instructions](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line) (make sure you don't select any scope just generate a simple token). If you are using [Github Actions](#configuring-github-actions-recommended) to deploy your portfolio you can skip this section.
 
-1. Create a file called .env in the root directory of your project, check the base file
+1. Create a file called .env in the root directory of your project (if not done already in section: [How To Use](#how-to-use))
 
-Note: Instead of creating a .env file, you can just run this command "cp env.example .env" inside the root directory
+Note: Configuring environment variables before deploying your portfolio is highly recommended as some components depend on API data. 
 
 ```bash
 - DeveloperFolio
--- a/README.md
+++ b/README.md
@@ -72,23 +72,23 @@ From your command line, clone and run developerFolio:
 
 ```bash
 # Clone this repository
-$ git clone https://github.com/saadpasta/developerFolio.git
+git clone https://github.com/saadpasta/developerFolio.git
 
 # Go into the repository
-$ cd developerFolio
+cd developerFolio
 
-# Setup environment variables
+# Setup default environment variables
 
 # For Linux
-$ cp env.example .env
+cp env.example .env
 # For Windows
-$ copy env.example .env
+copy env.example .env
 
 # Install dependencies
-$ npm install
+npm install
 
 # Start a local dev server
-$ npm start
+npm start
 ```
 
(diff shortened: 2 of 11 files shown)


kartikcho:
Let me know what you think about this decision and also check the test setup (don't have much experience with it). I followed a fix suggested from the official documentation [here](https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom)

README.md
```diff
@@ -77,20 +77,27 @@ $ git clone https://github.com/saadpasta/developerFolio.git
 # Go into the repository
 $ cd developerFolio
 
+# Setup environment variables
+
+# For Linux
+$ cp env.example .env
+# For Windows
+$ copy env.example .env
```
naveen521kk, line 85: the `$` here seems to misguide people, where people would just put the `$` in the terminal and get errors.

env.example
```diff
@@ -2,7 +2,10 @@
...
 // Do not change this file, adding your GITHUB TOKEN here!
 // Use cp or mv (like this: cp env.example .env), then edit .env with your GITHUB TOKEN.
-// IMPORTANT: Don't forget to add to update your .gitignore with .env (to avoid public share your key!)
+// IMPORTANT: Don't forget to add to update your .gitignore with .env (to avoid making your key public!)
 
 REACT_APP_GITHUB_TOKEN = "YOUR GITHUB TOKEN HERE"
 GITHUB_USERNAME = "YOUR GITHUB USERNAME HERE"
+// Set to true to fetch profile data from GitHub (remember to remove all components relying on GitHub data if set to false)
+// Ensure data type is string
+USE_GITHUB_DATA = "false"
```
naveen521kk, line 11: add a new line here.
Also, I think, since it's an example file I would set the default value to be `true`, what do you think?

env.example
```diff
@@ -2,7 +2,10 @@
 // This is only a .env example
 // Do not change this file, adding your GITHUB TOKEN here!
 // Use cp or mv (like this: cp env.example .env), then edit .env with your GITHUB TOKEN.
-// IMPORTANT: Don't forget to add to update your .gitignore with .env (to avoid public share your key!)
+// IMPORTANT: Don't forget to add to update your .gitignore with .env (to avoid making your key public!)
 
 REACT_APP_GITHUB_TOKEN = "YOUR GITHUB TOKEN HERE"
 GITHUB_USERNAME = "YOUR GITHUB USERNAME HERE"
+// Set to true to fetch profile data from GitHub (remember to remove all components relying on GitHub data if set to false)
+// Ensure data type is string
```
naveen521kk, line 10: this comment doesn't make sense to me since this is a `.env` file where things should be always as a string.

fetch.js
```diff
@@ -45,40 +53,41 @@ var data = JSON.stringify({
...
+    let data = "";
 
-  console.log(`statusCode: ${res.statusCode}`);
-  if (res.statusCode != 200) {
-    throw "The request to Github didn't suceed. Maybe check Github Token?";
-  }
+    console.log(`statusCode: ${res.statusCode}`);
+    if (res.statusCode !== 200) {
+      throw new Error(ERR.requestFailed);
+    }
```
naveen521kk, line 74: is this a typo? or an operator which exists in JS and I didn't know about?
```suggestion
    if (res.statusCode != 200) {
      throw new Error(ERR.requestFailed);
    }
```

naveen521kk (changes requested):
I don't know anything about test setup, so if things work then it is fine to have.

README.md
kartikcho: Good point, I'll remove them from all commands

env.example
kartikcho: I set it to false as default so that new users can setup dev environment without crashes and don't feel discouraged that it isn't working. 
However, keeping it to true is also an equally good decision as then it would throw errors on first run which would enforce people to setup these variables.

fetch.js
kartikcho: Not a typo, it's for not equal value or not equal type. Think of it as opposite of the '===' operator.

env.example
kartikcho: Fair point but I wanted to clear confusion for new folks about not thinking of it as a boolean and remove `""`. 
I'll remove the comment.

env.example
naveen521kk: > I set it to false as default so that new users can setup dev environment without crashes and don't feel discouraged that it isn't working.
On the first run, I think this file wouldn't be read at all. The file will only be read when it is changed to `.env`, which means the user has already made some efforts to make things work and it would be good to set it as true, and that won't be discouraging.

kartikcho:
Here's a screenshot of tests passing for reassurance that it works.

image

Not a pretty detailed functional test but it checks if all the components render without crashing and follow best practices etc.

naveen521kk:
> Here's a screenshot of tests passing for reassurance that it works.

Can we run the test in Github Actions?

env.example
kartikcho: Yep, set it to true but now we're assuming they've followed the new added steps to copy the env.example file before running the app locally so keeping that in mind too, setting it to default as true is good.

fetch.js
naveen521kk: looks like my knowledge about JS isn't that great

kartikcho:
We can but they're not super reliable, just indicating that nothing is broken. What do you think?

naveen521kk:
> We can but they're not super reliable, just indicating that nothing is broken.

Then, I think it's useless.

Also, in the vercel deploy, I couldn't see the Github Profile section? Also, did you set the env var for Github Actions deploy?

kartikcho:
> in the vercel deploy, I couldn't see the Github Profile section? Also, did you set the env var for Github Actions deploy?

I don't have the gh data variables configured on my local build, is that the reason why they're not showing up on the preview build?

naveen521kk:
> I don't have the gh data variables configured on my local build, is that the reason why they're not showing up on the preview build?

No.
I think @saadpasta have to set an env var called `USE_GITHUB_DATA` to `true` only then profile would be fetched according to recent changes. The same applies to the GH actions we have, you should edit them to set that env var.

.github/workflows/node.js.yml
```diff
@@ -3,6 +3,7 @@ env:
   CI: false
   GITHUB_USERNAME: ${{ github.repository_owner }}
   REACT_APP_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Use the generated personal access token and add it to repository secrets with the name OPEN_SOURCE_TOKEN
+  USE_GITHUB_DATA: ${{ secrets.USE_GITHUB_DATA }}
```
naveen521kk, line 6: why `${{ secrets.USE_GITHUB_DATA }}`?
```suggestion
  USE_GITHUB_DATA: "true"
```

kartikcho: Wouldn't this cause failing builds every time if the other variables aren't set?

naveen521kk: > Wouldn't this cause failing builds every time if the other variables aren't set?

Well, when running Github Actions, Github gives us a token to access their API and we use that for fetching details, the line
```
REACT_APP_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
means, it uses the one provided by Github itself, that comment near to it doesn't make sense though.

So, it means, the user doesn't need to do anything for Github Actions to work and this won't give failing builds.

kartikcho: Ah that does make sense, I thought it would fail requests if we don't provide other vars.

naveen521kk (approved):
lgtm

PR #362, opened 2021-05-03
Add carousel slider in work experience and achievement section

skgupta77159:
Hi, See I have added a carousel slider in the work experience and Achievement & certificate section. so that more experience and certificates can be added without just increasing the length of the portfolio. I found that if we add more achievements, the cards are increasing on the same page instead slider can do better for adding more than cards.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1155,6 +1155,29 @@
       "resolved": "https://registry.npmjs.org/@csstools/normalize.css/-/normalize.css-10.1.0.tgz",
       "integrity": "sha512-ij4wRiunFfaJxjB0BdrYHIH8FxBJpOwNPhhAcunlmPdXudL1WQV1qoP9un6JsEBAgQH+7UXyyjh0g7jTxXK6tg=="
     },
+    "@emotion/is-prop-valid": {
+      "version": "0.8.8",
+      "resolved": "https://registry.npmjs.org/@emotion/is-prop-valid/-/is-prop-valid-0.8.8.tgz",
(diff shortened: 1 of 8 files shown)


kartikcho (changes requested):
Thanks for opening this PR, this is a neat change that would make our existing components even better!

Took a brief look at the changes:

- Since our use case isn't complex, we can/should avoid adding a big package like `react-elastic-carousel` for adding a simple carousel. We could maybe implement our own carousel.
- There is no CSS styling for carousel buttons and carousel 'page' indicators in dark mode.
- ~The left button 'icon' is a bit off center, I think it might be a problem with the symbol and not with CSS.~ Edit: Looking at it again, it seems normal so you may discard this. 

image

- By default we don't need to keep dummy cards to show the carousel, we could add it in our docs that the components are flexible to hold multiple cards.
- Not sure how flexible the component is but the disabled icon on left (for scrolling to content that doesn't exist) looks ugly. We could either hide the button altogether in that case or set it to infinite scroll.

Let me know if you need any help or want a bit more clarification on my ideas.

kartikcho:
Hey @skgupta77159 are you still working on this?

saadpasta:
@kartikcho should we close this one also

kartikcho:
@saadpasta it looks like a good addition (maybe not the best implementation). We can either keep this open or make a new issue linking to this for reference because the portfolio should handle cases like having too many achievement cards without breaking or cluttering layout.

sadapasta:
@kartikcho I create a an issue linking to this PR

saadpasta:
@kartikcho I create a an issue linking to this PR

Issue #363, opened 2021-05-07
Convert Main.js into a functional component


SJellen:
## Summary
Convert Main.js into a functional component

### What's new?
Convert Main.js into a functional component

### Motivation
As far as I can tell, almost all the components are functional components except for Main.js. For consistency it would be better if Main.js was also a functional component. 

### Additional context
Main.js is intertwined with the theme changer so thats the main hangup on making a quick fix.

kartikcho:
Changed in https://github.com/saadpasta/developerFolio/commit/189dab9155929d360e21218b24a06d94fa2a5e9b

PR #365, opened 2021-05-09
Converted Main.js to a functional component

ashutosh1297:
Pull request in resolution of issue #363


src/containers/Main.js
```diff
@@ -18,51 +18,49 @@ import {StyleProvider} from "../contexts/StyleContext";
...
-    });
+    setIsDark(JSON.parse(localStorage.getItem("isDark")));
+  }, []);
+
+  const changeTheme = () => {
+    setIsDark(!isDark);
   };
+  useEffect(() => {
+    // Update local storage as soon as isDark changes
+    localStorage.setItem("isDark", isDark);
```
kartikcho, line 37: Why do we need a separate useEffect for this? Moving this localStorage call to the `changeTheme` function should handle this in `StyleProvider` ?

kartikcho (changes requested):
Looks good. Just need to address a couple of comments and fix the formatting check to get this in!

kartikcho:
@ashutosh1297 can you address the requested changes and then rebase #366 to reflect these changes.

ashutosh1297:
Ah sorry, my bad! I had typed the response but forgot to submit review. Please check my response and let me know how to proceed. Thanks! @kartikcho

src/containers/Main.js
kartikcho: You can abstract all of this away in a custom hook (probably should create a utils or hooks folder) called [`useLocalStorage()`](https://usehooks.com/useLocalStorage). This would also solve the problem of using the context API to get style state as we can switch to using only local storage calls.

I know this might add a bit more work than your original proposed changes but this seems like the most optimal and simplest solution going forward.

kartikcho:
It's all good @ashutosh1297, I've added a new comment with a more optimal solution than the one we have been using. This should also slightly improve performance of our app.

ashutosh1297:
@kartikcho I moved the whole localStorage logic to a custom hook, checkout the new commit for updates. 

Keeping the context API to share theme state through the component tree seems like a better idea than using localStorage calls everywhere, can you elaborate on how you were planning of getting rid of the context API?

PS - sorry about the delay in replies, I am a little caught up with work but I will try and wrap this up soon

kartikcho (approved):
No worries, thanks for the update!
In hindsight, I thought we could move away from Context altogether and check for the localStorage key where ever we need the dark mode flag. Thinking about it now, it might cause needless overhead. 
Lets scratch that idea and stick with Context calls.

kartikcho:
Hey @ashutosh1297, could you run prettier on this branch to fix the checks?

naveen521kk:
It's very bad that prettier doesn't show the file diff. Let me see if I can fix it.

naveen521kk:
> It's very bad that prettier doesn't show the file diff. Let me see if I can fix it.

Can we set up pre-commit https://prettier.io/docs/en/precommit.html#option-3-pre-commithttpsgithubcompre-commitpre-commit? I usually use it other python project it works and I can setup Github Actions to run that and output the diff. Tbh, I can't find a better idea looking at the docs.

kartikcho:
I have used Husky in the past for adding a pre commit hook but it's a hit or miss honestly, it only works correctly for some git users.
https://pre-commit.ci/ looks promising and if we can set it up, that'd be great too. Anyway, @naveen521kk we should log the errors from the prettier gh action.
Also @ashutosh1297, the check is still failing 😅

naveen521kk:
> https://pre-commit.ci/ looks promising and if we can set it up, that'd be great too.

Should be simple enough that @saadpasta should install a Github App as instructed there. I will make a PR for adding pre-commit config then :).

saadpasta:
@naveen521kk  I have added pre-commit cli app

naveen521kk:
>  I have added pre-commit cli app

Thanks, I will set up pre-commit in another PR.

kartikcho:
pre-commit.ci autofix

kartikcho:
@naveen521kk shouldn't this just run the prettier command in a new commit on this PR? So why is it failing? 
Their logs are anything but helpful.

naveen521kk:
pre-commit.ci autofix

naveen521kk:
Hmm, this looks confusing.

ashutosh1297:
@kartikcho I am not sure why it's failing, I ran the prettier script from package.json and it showed that everything's fixed. Are there more steps involved?

ashutosh1297:
> Also @ashutosh1297, the check is still failing 😅

@kartikcho I am not sure why it's failing, I ran the prettier script from package.json and it showed that everything's fixed. Are there more steps involved?

naveen521kk:
> @kartikcho I am not sure why it's failing, I ran the prettier script from package.json and it showed that everything's fixed. Are there more steps involved?

Can you rebase with the master and then run `npm run format`? Maybe it fixes this.

kartikcho:
I'll probably take a look at the tests this weekend but for now, try doing what Naveen suggested.

saadpasta:
@kartikcho are you looking into this.

kartikcho:
Sorry, I've been busy and this totally skipped my mind. Will try to take a look whenever possible.

kartikcho:
Hey @ashutosh1297, rebase from master and this should be green.

saadpasta (approved):
LGTM  👍  
@ashutosh1297  please rebase from master

kartikcho:
Hey @ashutosh1297 are you still working on this?

PR #368, opened 2021-05-22
Declare all files have LF line endings

naveen521kk:
should fix prettier tests in https://github.com/saadpasta/developerFolio/pull/355


kartikcho (approved):
Seems like this should fix the issue. What's up with the static assets being picked up by git diff?

naveen521kk:
> What's up with the static assets being picked up by git diff?

no idea.

PR #372, opened 2021-06-04
Add pre-commit config

naveen521kk:
As discussed in https://github.com/saadpasta/developerFolio/pull/365#issuecomment-853227996


.pre-commit-config.yaml
```diff
@@ -0,0 +1,8 @@
+repos:
```
kartikcho, line 1: Shouldn't we also add config for CI stuff like [here](https://pre-commit.ci/#configuration) to setup commit msg etc.

kartikcho (approved):
The config looks good, what's up with pre-commit failing on itself?

.pre-commit-config.yaml
naveen521kk: I think that isn't required, it uses a default commit message which will make sense.

naveen521kk:
pre-commit.ci autofix

PR #379, opened 2021-06-20
Prettier version bump

kartikcho:
Fixed a formatting error in master, this should also fix the failing check in other PRs like #365.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3098,14 +3098,37 @@
       }
     },
     "browserslist": {
-      "version": "4.14.5",
-      "resolved": "https://registry.npmjs.org/browserslist/-/browserslist-4.14.5.tgz",
-      "integrity": "sha512-Z+vsCZIvCBvqLoYkBFTwEYH3v5MCQbsAjp50ERycpOjnPmolg1Gjy4+KaWWpm8QOJt9GHkhdqAl14NpCX73CWA==",
+      "version": "4.16.6",
+      "resolved": "https://registry.npmjs.org/browserslist/-/browserslist-4.16.6.tgz",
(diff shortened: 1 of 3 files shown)


naveen521kk:
Hmm, why does CI fail here?

kartikcho:
Okay I think I figured out what's wrong with the prettier checks, it's only looking at master everytime as mentioned in the workflow config.

Running the same check locally gives me green, so this is definitely checking the current state of master everytime which is failing as of now.

image

naveen521kk:
> Okay I think I figured out what's wrong with the prettier checks, it's only looking at master everytime as mentioned in the workflow config.

Do you mean it checks out master each time it runs the checks (I mean the Github Action)?

kartikcho:
Yes, master has formatting errors (which this PR addresses) and this is why formatting fails in every PR. And this action would also show green on PRs with formatting errors if the master is green.

naveen521kk:
> Yes, master has formatting errors (which this PR addresses) and this is why formatting fails in every PR. And this action would also show green on PRs with formatting errors if the master is green.

No, I don't think so. It checks out this branch (see the logs https://github.com/saadpasta/developerFolio/pull/379/checks?check_run_id=2868560528#step:2:40) correctly. Maybe it's about the prettier version, CI install's `2.3.1` and in `package-lock.json` it is `2.3.0` https://github.com/saadpasta/developerFolio/blob/master/package-lock.json#L10977. Can you try updating and locally and then run `npm run format`?

kartikcho:
Good catch, you're right, it was a result of version mismatch

naveen521kk:
I think we should set up the depandabot to at least upgrade this.

naveen521kk (approved):
lgtm

PR #410, opened 2021-09-15
Allow users to setup SEO, tweak metadata and package files

### LINKED ISSUE #399: Allow users to tweak SEO
Our current application has pretty much non-existent SEO and no option to help users tweak it either. 
We should look into making this accessible for new folks with proper documentation explaining its importance and use too.

kartikcho:
Resolves #399 

Using [react-helmet](https://github.com/nfl/react-helmet), users can now setup their portfolio's SEO and other Open Graph meta information from `portfolio.js`.

This PR also introduces changes to set default metadata to 'DeveloperFolio' instead of 'Saad Pasta Portfolio'. This would solve the problem for old and new users who have not setup SEO, to not have Saad's name on every fork of the app. 
Let me know what you think about this change and if @saadpasta would like to extend this further by making the [preview deployment](https://developerfolio.js.org/) use filler data (use strings like FirstName LastName instead of Saad Pasta) which can help beginner users more imo.

(To test locally if the meta/og changes are working, I used an SEO extension and would suggest the same for reviewers)


sadapasta:
@kartikcho may be good to include this in the readme.MD also

saadpasta (commented):
Maybe include this work in readme.md

saadpasta:
@kartikcho I am open to change my name to something else, but not to First Name or Last Name but good with some random name like Jon, or something

naveen521kk:
Hmm, I am slightly confused, does react-helmet adds the metadata while building the site itself or does it add it in client-side? If it's client-side it's not worth the effort since this will not show up on social media pages(like Twitter or discord) and probably search engines could ignore it.

kartikcho:
> 
> 
> @kartikcho I am open to change my name to something else, but not to First Name or Last Name but good with some random name like Jon, or something

I'll use el clásico: 'John Doe', unless you have any preferences

kartikcho:
> does react-helmet adds the metadata while building the site itself or does it add it in client-side? If it's client-side it's not worth the effort since this will not show up on social media pages(like Twitter or discord) and probably search engines could ignore it.

I was aware of this only being a client-side solution but I think some search engines like Google can index it now. The only other options I see is adding docs to directly tweak the HTML (simple but the user will have to touch another file) or maybe running another workflow to set it up during build time (overengineering?)

naveen521kk:
> maybe running another workflow to set it up during build time (overengineering?)

Probably, that's better than having people complaing the metadata not shown on social media? I think we can use react-helmet but instead of client side it should be done before deploying, like how the profile data from github is saved. Or we can simply ask users to edit the HTML file.

kartikcho:
Another workflow to touch the html file sounds more trouble than it's worth. 
Let's just stick with the HTML solution, simpler to understand for new people too.

kartikcho:
Didn't remove Saad's name from the landing (didn't feel the same without it ❤). The new changes are much simpler and don't require a new dependency. Take a look @naveen521kk and lemme know if this works.

kartikcho:
OpenGraph test: works ✅
image

naveen521kk (approved):
lgtm

Issue #212, opened 2020-07-15
Add custom error messages for console


kartikcho:
We should add custom error messages to the console for components that depend on third party data like Twitter, Github projects and podcasts.
This would clear confusion for users and visitors both on why some components are missing. 

Related to #165

SamratDe:
Hi. I would like to work on this issue.

kartikcho:
Alright, assigning this to you @SamratDe !

SamratDe:
@kartikcho I am fixing the warning messages in console first, because I think if there are too many messages in the console, users will have difficult time in finding the error message for components not loading. Is it ok to do so?
Most of the warning of messages are not adding key prop or misspelling prop or classNames etc.

kartikcho:
@SamratDe they don't show up in the production deployment so custom messages would still be on top. #138 has fixed these errors but we haven't merged it in yet because of the dark mode PR. 
Maybe you can follow up in another PR if there are still any messages left after that PR gets in, I appreciate the help!

SamratDe:
Ok. Then I will only concentrate on the error messages in the components.

SamratDe:
@kartikcho I have created a PR.

MuhammadHasham23:
@kartikcho I have made a [PR](https://github.com/saadpasta/developerFolio/pull/270) that will close this issue, feel free to look into it and let me know if that is fine.
Thanks.

Rispectech:
@kartikcho  if this issue is open then I would like to take it

kartikcho:
Sure @Rispectech! You can read the discussion above before opening a PR and ask if you have any questions.

I would suggest you to do one issue before taking more at once to keep other tasks available for other peeps too. (also easier to work this way)

Rispectech:
Thanks a lot @kartikcho . yeah u are right sorry got a bit ahead of myself

What I got from the above discussion, u need an error message on the console given the information from the external APIs is null. so sure it looks good my suggestion is we can set some default values if we don't get the information along with the error messages on the console. please tell me whether it is good enough. 
also, can I get your discord id or something? I would be extremely delighted if we can talk and can clear any confusion in case

kartikcho:
Yes error messages on the console would be best. Also, I think default values would defeat the purpose of adding errors since we want them to blow up in the user's face to tell them that stuff isn't working. Just make sure to throw informative errors that would help users find out what's wrong.

[...]

PR #430, opened 2021-10-02
console-err-msg

Rispectech:
In order to prevent redundancy in each component, I created a new js file (errorfunc.js) which contains all the error functions I used apart from occasional exceptions where it couldn't be used 

in errorfunc.js , checkMissingValueObj checks whether the values are present or not

### DIFF
--- a/src/components/educationCard/EducationCard.js
+++ b/src/components/educationCard/EducationCard.js
@@ -2,11 +2,15 @@ import React, {createRef, useContext} from "react";
 import {Fade, Slide} from "react-reveal";
 import "./EducationCard.scss";
 import StyleContext from "../../contexts/StyleContext";
+import {checkMissingValuesObj} from "../errorfunc";
 
 export default function EducationCard({school}) {
   const imgRef = createRef();
 
+  checkMissingValuesObj(school, school.name, "Education");
+
   const GetDescBullets = ({descBullets}) => {
+    if (!descBullets) return null;
     return descBullets
       ? descBullets.map((item, i) => (
           <li key={i} className="subTitle">
@@ -16,6 +20,9 @@ export default function EducationCard({school}) {
       : null;
   };
   const {isDark} = useContext(StyleContext);
+
+  if (!school.logo)
+    console.log(`Image of schoolName ${school.name} is not available`);
   return (
     <div>
       <Fade left duration={1000}>
--- a/src/components/errorfunc.js
+++ b/src/components/errorfunc.js
@@ -1,0 +1,48 @@
+//function to check whether the url is correct
+
+function isValidHttpUrl(string) {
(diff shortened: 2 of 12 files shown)


src/components/educationCard/EducationCard.js
```diff
@@ -16,6 +20,9 @@ export default function EducationCard({school}) {
       : null;
   };
   const {isDark} = useContext(StyleContext);
+
+  if (!school.logo)
+    console.log(`Image of schoolName ${school.name} is not available`);
```
kartikcho, line 25: We can keep this in as it seems helpful but change it to `console.error('Image of ${school.name} is missing in education section');`

src/components/errorfunc.js
```diff
@@ -0,0 +1,48 @@
+//function to check whether the url is correct
```
kartikcho, line 1: You'll probably need to scrap all of this code and it's usage in other components as it isn't what is required currently :(

kartikcho (changes requested):
I took a quick look over the changes and it seems you understood the issues incorrectly. The issue #212 says to add error messages for only third party components, i.e GitHub repos, Twitter embed, podcasts and the GitHub contact components.

Throwing errors for missing fields would be counter intuitive as we also want users to have the option to keep fields blank if they wish to do so. Lots of errors in the console blowing up in people's face will make them feel they're doing something wrong when it's just an empty field.

Also looking at the implementation, you're using logs instead of errors which isn't the best way to warn someone.

Rispectech:
> I took a quick look over the changes and it seems you understood the issues incorrectly. The issue #212 says to add error messages for only third party components, i.e GitHub repos, Twitter embed, podcasts and the GitHub contact components.
> 
> Throwing errors for missing fields would be counter intuitive as we also want users to have the option to keep fields blank if they wish to do so. Lots of errors in the console blowing up in people's face will make them feel they're doing something wrong when it's just an empty field.
> 
> Also looking at the implementation, you're using logs instead of errors which isn't the best way to warn someone.

okay, so you want the error using throw error instead of log.right will do this one 
second right using the error function for all the components intuitive so fix this as well 
and how about editing the error function to throw error instead of console message for missing  values in repos card
and a doubt can u please tell how the rest of components fetch data ( Twitter embeds , podcast )

Rispectech:
i have converted the console messages in third party components to error messages and removed them from the rest 
errorfunc.js deleted as well
please check and review the changes
thanks

Rispectech:
sorry @kartikcho for being such a pain making you request changes again and again 
I got confused with throw and console when you last commented so made these mistakes so sorry for that 
I have made the changes so please review and check 
thanks

kartikcho:
@allcontributors add @Rispectech for code

PR #435, opened 2021-10-12
 Added carousel slider in work experience and achievement section issues: #415

S-codes14:
this is what @saadpasta  said "We can add a Carousel slider in the work experience and Achievement & certificate section so that more experience and certificates can be added without just increasing the length of the portfolio. I found that if we add more achievements, the cards are increasing on the same page instead of the slider can do better for adding more than cards."
[here](https://github.com/saadpasta/developerFolio/issues/415)

I tried to come up with a solution and it works

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,10 @@ const twitterDetails = { ... }
 
 For adding emoji 😃 into the texts in `Portfolio.js`, use the `emoji()` function and pass the text you need as an argument. This would help in keeping emojis compatible across different browsers and platforms.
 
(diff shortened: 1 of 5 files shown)


saadpasta:
Closing this PR due to no response

S-codes14:
I was going to respond it is just that you guys are unclear on what type of carousel you want

sadapasta:
I don't think so there should be carousel for this maybe better to fix this in just one or 2 grid maybe

PR #446, opened 2021-11-15
Fix fallback for medium blogs

CanciuCostin:
Setting displayMediumBlogs to true breaks since there are no fetched blogs.


naveen521kk:
Can you rebase on the current master? Also, by default if the JSON file is unavailable, it should fall back to hardcoded value so that doesn't work. You can do something like https://github.com/saadpasta/developerFolio/commit/e0913936cd9140de8422b88c26ad8ccd029d62f9

naveen521kk (approved):

PR #457, opened 2022-01-15
Changed font size from px to rem

### LINKED ISSUE #456: Switch font sizes to using 'rem' units instead of pixels or 'em'
## Summary
We should switch to using 'rem' units instead of the inconsistent pixel values for improving accessibility and following accessible web design best practices.

### What's new?
The font sizes should switch to 'rem' but the final design should remain almost completely similar to before.
For conversion reference, 16px = 1rem.

### Motivation
Improving accessibility and following accessible web design best practices.

SpectralGT:
Fixes #456 

### Changes Made
I change the font size form px to rem.

### What Happened
The size is same but the unit is different.

### Are there any Mistakes
Visually every thing looks the same as stated in the issue.

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -17,7 +17,7 @@
   flex-direction: column;
   align-items: center;
   justify-content: center;
-  font-size: calc(10px + 2vmin);
+  font-size: calc(0.625rem + 2vmin);
   color: $textColorDark;
 }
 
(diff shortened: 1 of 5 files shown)


saadpasta (approved):
LGTM. I am not big CSS guys 😄

saadpasta:
@allcontributor please add @SpectralGT for code

saadpasta:
@allcontributors please add @SpectralGT for code

PR #459, opened 2022-01-22
Add .vs/ to .gitignore

Sinasun:


saadpasta:
I am not sure if this is needed right now , but thanks for for contributing ❤️

PR #464, opened 2022-02-18
Medium cards are not navigating.

techiesalman:
When a user clicks on the Medium blog card, instead of navigating to the medium page it stays over there only. So instead of mapping blog.link, code is mapped with blog.url. Because of this, no proper URL gets mapped to blog cards.

### DIFF
--- a/src/containers/blogs/Blogs.js
+++ b/src/containers/blogs/Blogs.js
@@ -83,7 +83,7 @@ export default function Blogs() {
                       key={i}
                       isDark={isDark}
                       blog={{
-                        url: blog.url,
+                        url: blog.link,
                         title: blog.title,
                         description: extractTextContent(blog.content)
                       }}


kartikcho:
Huh, that's very weird. 
If you look at the portfolio manifest, blog has a [url](https://github.com/saadpasta/developerFolio/blob/master/src/portfolio.js#L303) key but it isn't working in the deployed version (and is working in yours). Can you set the blog object key to link and check if it still works with your code @techiesalman? 
More surprised no one ever noticed this before 😅

sadapasta:
@kartikcho I think it's better to change it in the component rather in the portfolio.js file

sadapasta (approved):
LGTM 👍 Good Catch

saadpasta (approved):

saadpasta:
Always keep work and personal profile on different browser

PR #501, opened 2022-06-25
docs: add USE_GITHUB_DATA = "true" to the .env file in documentation

Ishan-001:
Currently, github data for one's profile isn't fetched unless we add `USE_GITHUB_DATA = "true"` to the `.env` file. Maybe adding this line in the documentation might be helpful.


kartikcho (approved):
Good catch, thank you!

kartikcho:
@allcontributors please add @Ishan-001 for docs

PR #512, opened 2022-08-27
Removed duplicated component, fixed hardcoded Image alt

slyapustin:
Hey guys!

I really like your project, here a couple of small fixes:
- Removed duplicated component with a typo in the name.
- Fixed hardcoded Image Alt in the Achievements.

Thanks!


src/components/achievementCard/AchievementCard.js
```diff
@@ -14,7 +14,11 @@ export default function AchievementCard({cardInfo, isDark}) {
   return (
     <div className={isDark ? "dark-mode certificate-card" : "certificate-card"}>
       <div className="certificate-image-div">
-        <img src={cardInfo.image} alt="PWA" className="card-image"></img>
+        <img
+          src={cardInfo.image}
+          alt={cardInfo.imageAlt}
```
kartikcho, line 19: Could you add a basic fallback alt text just in case someone doesn't configure it?
Like: `alt={cardInfo.imageAlt || "card thumbnail"}`

src/components/achievementCard/AchivementCard.js
```diff
@@ -1,32 +0,0 @@
-import React from "react";
-
-export default function AchivementCard({cardInfo}) {
```
kartikcho, line 3: oh, didn't know this existed lol. Thanks!

kartikcho (approved):
Great PR, just need to address one comment

src/components/achievementCard/AchievementCard.js
slyapustin: I don't think you need a default. 
If you did not provide an `imageAlt` it will just not pass an `alt` attribute to an `` at all, which looks right to me.

slyapustin: @kartikcho Actually, I think you are right - it's better to have a default value for the alt. I just updated the PR.

kartikcho:
@allcontributors please add @slyapustin for code

Issue #490, opened 2022-05-23
Same Height For the OpenSource Project Cards


saiteja13427:
**Describe the bug**
Currently the height of the opensource project cards varies depending on the project content. I think we should try making the height same. 

**To Reproduce**
Its visible on the homepage

**Screenshots**
![image](https://user-images.githubusercontent.com/40917760/169863166-35382581-08ff-48f7-b0ce-dafb3d54d0d6.png)

**Note:** I would like to take this up, please assign it to me.

kartikcho:
Sorry for the late reply, I noticed this too. Thanks for taking this up, assigning it to you @saiteja13427

saiteja13427:
Thanks @kartikcho, Will correct it and raise a PR shortly

PR #516, opened 2022-08-28
Equal Opensource Card Sizes

saiteja13427:
Makes all the opensource project card sizes equal.

### Changes
1. Changed the open-source project cards from the grid layout to flex.
2. Made the heights to be equal.
3. Also we will cut the name of the project if it is too long.

Addresses Issue #490 

@kartikcho Do have a look!

**Note:** Seems like the vercel auto deployment is not configured to show up the opensource cards, so you will have to test it locally i guess.

### DIFF
--- a/src/components/githubRepoCard/GithubRepoCard.js
+++ b/src/components/githubRepoCard/GithubRepoCard.js
@@ -13,85 +13,83 @@ export default function GithubRepoCard({repo, isDark}) {
   }
 
   return (
-    <div
-      className={isDark ? "dark-card-mode repo-card-div" : "repo-card-div"}
-    >
-    <Fade bottom duration={1000} distance="20px">
-      <div>
-        <div
-          key={repo.node.id}
-          onClick={() => openUrlInNewTab(repo.node.url, repo.node.name)}
-        >
-          <div className="repo-name-div">
-            <svg
-              aria-hidden="true"
-              className="octicon repo-svg"
-              height="20"
-              role="img"
-              viewBox="0 0 12 16"
-              width="14"
-            >
-              <path
-                fill-rule="evenodd"
-                d="M4 9H3V8h1v1zm0-3H3v1h1V6zm0-2H3v1h1V4zm0-2H3v1h1V2zm8-1v12c0 .55-.45 1-1 1H6v2l-1.5-1.5L3 16v-2H1c-.55 0-1-.45-1-1V1c0-.55.45-1 1-1h10c.55 0 1 .45 1 1zm-1 10H1v2h2v-1h3v1h5v-2zm0-10H2v9h9V1z"
-              ></path>
-            </svg>
-            <p className="repo-name">{repo.node.name}</p>
-          </div>
-          <p className="repo-description">{repo.node.description}</p>
-          <div className="repo-stats">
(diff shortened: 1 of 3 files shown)


src/components/githubRepoCard/GithubRepoCard.js
```diff
@@ -13,82 +13,83 @@ export default function GithubRepoCard({repo, isDark}) {
...
-              ></path>
-            </svg>
-            <p className="repo-name">{repo.node.name}</p>
-          </div>
-          <p className="repo-description">{repo.node.description}</p>
-          <div className="repo-stats">
-            <div className="repo-left-stat">
-              {repo.node.primaryLanguage !== null && (
+    <div className={isDark ? "dark-card-mode repo-card-div" : "repo-card-div"}>
+      <Fade bottom duration={1000} distance="20px">
```
kartikcho, line 17: Why is 'Fade' not the parent anymore? 
If there's no particular reason for the reorder, can we revert? I'd like to keep the same pattern across components for consistency

src/components/githubRepoCard/GithubRepoCard.scss
```diff
@@ -54,8 +68,6 @@
 }
 
 .repo-name {
-  white-space: nowrap;
-  text-overflow: ellipsis;
```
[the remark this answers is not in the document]
saiteja13427: Yeah, because of changing to flex, ellipsis will not work. If we want ellipsis, then we need to make some more changes.

src/components/githubRepoCard/GithubRepoCard.js
saiteja13427: Ya in that case I think we can revert this, shall I make it the parent again and push a commit?

kartikcho: Yeah that'd be great for consistency

src/components/githubRepoCard/GithubRepoCard.scss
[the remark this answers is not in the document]
kartikcho: I tried it and it breaks the layout into a long single card (which is probably how it was before your changes too). 
We can change the layout container to use grid or add a max width to the card so it doesn't extend into a single long card. This would also solve the ellipsis problem as the container would now have a specified width so the p tag can just be `width: 100%` now

image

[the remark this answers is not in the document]
saiteja13427: Sure I will do that and push!

src/components/githubRepoCard/GithubRepoCard.js
saiteja13427: Alright!

kartikcho:
Are you still working on this @saiteja13427 ?

saiteja13427:
@kartikcho Yes, I was stuck with some other work. Will push the changes today or tomorrow for sure

kartikcho:
Sure np, thanks for the heads up!

saadpasta:
@saiteja13427 Are you still working on it or should I close it.

saiteja13427:
@saadpasta I think we can close this PR for now. I will raise a new PR with the changes in the suggested way!

Issue #521, opened 2022-09-07
Education Logo Looks Weird If User Doesn't give the Logo in Portfolio.js


saiteja13427:
**Describe the bug**
If the developer doesn't give the logo field in education, the education card turns out to be weird.

**To Reproduce**
1. Remove the logo from any one of the education objects in portfolio.js
2. And you can see something like this

![image](https://user-images.githubusercontent.com/40917760/188932952-390a3937-0312-42fc-9899-255d0500449d.png)

**Expected behavior**
We should check for the presence of the logo and only then render it.

**Is this responsiveness Issue**
NO

@kartikcho If you think it is worth correcting, then I will do it.

saiteja13427:
Yeah, raised a PR #522 @kartikcho, do check and merge it.

saiteja13427:
As the PR is merged, the issue now stands closed!

PR #522, opened 2022-09-07
Check for School Logo and Render

saiteja13427:
Just added a simple check for school logo.

### Reason
Without the check if the user doesn't give a logo, then it would render something weird which is not good. As discussed in #521 

### Before
image

### Now
image


saiteja13427:
@kartikcho Did you get a chance to look at this PR?

kartikcho (approved):
Simple fix, neat!

PR #529, opened 2022-10-01
Added repo size conversion to github repo cards.

### LINKED ISSUE #525: Add repo size conversion to github repo cards
Cards currently only show repo sizes in kb. This looks ugly when a repo size is in mb or greater.

We should add size conversion and format the output appropriately

Vinitvh:
Added repo size conversion to GitHub repo cards closes #525

### DIFF
--- a/src/components/githubRepoCard/GithubRepoCard.js
+++ b/src/components/githubRepoCard/GithubRepoCard.js
@@ -1,7 +1,7 @@
 import React from "react";
 import "./GithubRepoCard.scss";
 import {Fade} from "react-reveal";
-import {format} from "./sizeConversion";
+import {formatFileSizeDisplay} from "./sizeConversion";
 
 export default function GithubRepoCard({repo, isDark}) {
   function openUrlInNewTab(url, name) {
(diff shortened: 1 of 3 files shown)


src/components/githubRepoCard/GithubRepoCard.js
```diff
@@ -1,6 +1,7 @@
 import React from "react";
 import "./GithubRepoCard.scss";
 import {Fade} from "react-reveal";
+import {format} from "./sizeConversion";
```
kartikcho, line 4: Can we rename this to something descriptive like `formatFileSizeDisplay`?

src/components/githubRepoCard/sizeConversion.js
```diff
@@ -0,0 +1,6 @@
+export const format = value => {
```
kartikcho, line 1: You could move this function into a `utils.js` file in `/src`. That way we can add more functions in one place or share it amongst files if needed

kartikcho (commented):
Need to address some changes and then we're good to merge this!

src/components/githubRepoCard/sizeConversion.js
Vinitvh: Okay will make necessary changes.

kartikcho (approved):
Great PR! Thanks

kartikcho:
@allcontributors please add @Vinitvh for code

Issue #540, opened 2022-10-04
Upgrade Node as v12.x has been deprecated by Vercel


kartikcho:
Looks like vercel deprecated node 12 according to this [build log](https://vercel.com/saadpasta/developer-folio/7Smu8x9pnZKARHwAmsB1fu8cYaeT).
Should upgrade it to either the latest or nearest stable version without breaking too many things

kartikcho:
@saadpasta this is urgent
I can't config this. You'll have to fix this from the project settings on vercel. Until then no previews and deploys would work

SR-Sunny-Raj:
Please I request you to fix it asap as most of the dependencies are not getting downloaded and also the build is getting failed.

sadapasta:
@kartikcho updated to 14 in vercel can you try again

kartikcho:
@saadpasta could you add me to the project on Vercel? I think I'll need to manually trigger redeploys

kartikcho:
or @sadapasta, whichever one you're active on now :P

PR #544, opened 2022-10-05
feat: add fading Lottie

vatana7:
# Description

I've seen that on initial load of the page, Lottie seem to pop up suddenly so I made some adjustment for a fade in effect for Lottie to increase better UI.

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

==Before==
before

==After==
after


kartikcho:
@vatana7 could you upload a full video (I think you can upload videos now) of the change in action? The gifs aren't clear and layout in the 2nd one looks broken.

Ideally the vercel deployment would've helped out but waiting on @saadpasta to fix #540

vatana7:
I don't have any access to screen recorder at the moment but if you're curious I made some change with the div that cover the lottie to add in a little fade in effect on initial load to look better overall. If you think it's helpful please merge the code 🤟🏻. 
Also after #540 my effect would still be relevant as well

src/containers/greeting/Greeting.js
```diff
@@ -52,7 +52,9 @@ export default function Greeting() {
           </div>
           <div className="greeting-image-div">
             {illustration.animated ? (
-              <DisplayLottie animationData={landingPerson} />
+              <div className="fade-in">
```
kartikcho, line 55: We could use `<Fade bottom duration={1000} distance="40px">` from line 19 so both the text and the lottie appears with the same effect together

kartikcho: Could you check how that looks and if they fade in from bottom together?

kartikcho: A new commit would also trigger a redeploy

vatana7: Sure! I'll take a look and get back to you!

vatana7:
```javascript
<Fade bottom duration={5500} distance="60px">
       <DisplayLottie animationData={landingPerson} />
</Fade>
```
This is the adjustment I made as you suggested. Please have a look @kartikcho 

Kapture 2022-10-18 at 09 59 21

saadpasta:
@vatana7 Can we make the text and image show up at the same time currently the image slides in after the text

abdullahnasir4:
@saadpasta Can you share the url for landingperson.json from lottiefiles.

PR #549, opened 2022-10-31
edu-placeholder-img added & education section minor ui fix

PeerSyedHussain:
Hy team,

**In education section :**

- If any school/college logo is not available to show in the UI, we can use the placeholder image which represents school/college generally

- If any education doesn't have any points to be add in **descBullets** in **educationInfo** object ...we can use it as empty array so in UI it does not show like dotted

In my case I don't have my school logo and also I not has point to be added in descBullets. So please check from your end if its fine this PR, kindly accept it.

Thanks & Regards
Peer Syed Hussain S


saadpasta (commented):
I am not sure if this is useful or not. If someone school don't have a logo they can use whichever they like and I also don't the school placeholder image being used else where

PeerSyedHussain:
Ok I got it bro... In my case, I preferred to use exact logo or else placeholder image for that related topic...

If it's needed in future we can move this PR, or else will just leave as it is...No pb.

Thanks

src/components/educationCard/EducationCard.js
```diff
@@ -7,12 +7,14 @@ export default function EducationCard({school}) {
   const imgRef = createRef();
 
   const GetDescBullets = ({descBullets}) => {
-    return descBullets
-      ? descBullets.map((item, i) => (
-          <li key={i} className="subTitle">
-            {item}
-          </li>
-        ))
+    return descBullets && descBullets.length > 0
```
tasawar-hussain, line 10: descBullets?.length > 0

src/components/educationCard/EducationCard.js
```diff
@@ -7,12 +7,14 @@ export default function EducationCard({school}) {
...
+      ? descBullets.map((item, i) =>
+          item !== "" ? (
+            <li key={i} className="subTitle">
```
tasawar-hussain, line 13: not recommended to use the index as a key

tasawar-hussain, line 13:
```suggestion
      ? descBullets.filter(item => item !== "").map((item, i) =>
             (
            <li key={i} className="subTitle">
```

tasawar-hussain (changes requested):

Issue #579, opened 2023-01-11
Move StartupProject Section to Next Project


saadpasta:
https://github.com/saadpasta/developerFolio/tree/master/src/containers/StartupProjects

IAmTamal:
Can i work on this ? 
I guess this work should start only after #571 is done ?

IAmTamal:
- Also for achievement section even though there's a hover effect - the cursor is not of style `pointer` - we can have that little addition too.
- The buttons are kinda hard to read for mobiles : 
![image](https://user-images.githubusercontent.com/72851613/215036067-ddc97c24-ea45-45ef-b52e-809939c3b6d7.png)

saadpasta:
@IAmTamal would you like to do that I can assign you

saadpasta:
Issue #571 is completed

saadpasta:
Assigned to you

IAmTamal:
> Assigned to you

The branch is `main-next` right ? So i make a PR to that branch after i make changes right ?

saadpasta:
Yes. `main-next` @IAmTamal

PR #594, opened 2023-02-01
Move Blog section to nextjs project

### LINKED ISSUE #581: Move Blogs Section to Next Project
https://github.com/saadpasta/developerFolio/tree/master/src/containers/blogs

muneebahmedayub:
# Description

Blog section is moved to nextjs project. Some of the things I would like to mention:
- `app` directory is being used instead of `pages` to follow the new nextjs project structure and to use server components.
- as server components works only inside app directory so the `components` and `containers` folders are moved to `app` directory.
- folder name `(components)` is used instead of `components` because it is recommended way to structure folders inside app directory and also it doesn't make any route.
- local `.scss` files are converted to `.module.scss`.
- removed unnecessary classes
- `BlogCardClient` is used as a client component to handle only click event everything else is on server.

Fixes #581 

## Type of change

- [x ] New feature (non-breaking change which adds functionality)


muneebahmedayub:
@saadpasta I think vercel is using node version below 16 and with that it is not able to deploy

saadpasta (commented):
Do we need the brackets around components and components name ? Is that nessacry I don't think so. So I guess we can remove that

saadpasta:
Everything as look great

muneebahmedayub:
Thanks for noticing, after some research, I think we don't even need components and containers directory inside ```app``` directory they can be inside ```src```, I will fix this in next commit.

saadpasta:
@muneebahmedayub Sure man that's awesome let's do that ASAP I really like this and would wanna continue with your approach

muneebahmedayub:
Now ```app``` directory only contains routing information and all other files are moved from ```app``` to ```src```. And also removed paranthesis around components name.

IAmTamal:
Oh wow these are great additions, i think i'll make more changes and then make the PR for #579 after this one is merged !

src/app/layout.jsx
```diff
@@ -0,0 +1,14 @@
+import "@/styles/globals.scss";
+
+export default function RootLayout({children}) {
+  return (
+    <html lang="en">
+      {/*
+        <head /> will contain the components returned by the nearest parent
+        head.jsx. Find out more at https://beta.nextjs.org/docs/api-reference/file-conventions/head
```
sadapasta, line 8: Please remove this

src/app/page.jsx
```diff
@@ -0,0 +1,10 @@
+import Blogs from "../containers/Blogs/Blogs";
+
+export default async function Home() {
+  return (
+    <>
+      <div>DeveloperFolio Migration to be done</div>
```
sadapasta, line 6: Remove this div blog

src/containers/Blogs/Blogs.jsx
```diff
@@ -0,0 +1,75 @@
...
+                      url: blog.link,
+                      title: blog.title,
+                      description: extractTextContent(blog.content)
+                    }}
+                  />
+                );
+              })}
+        </div>
+      </div>
+    </div>
```
sadapasta, line 71: Can we try refactoring the part a bit? Maybe move the API call in the folder called client

IAmTamal: > Can we try refactoring the part a bit? Maybe move the API call in the folder called client

Won't it be better if the stuffs are fetched server side itself ? 
Also i think `cache: "force-cache"` is set by default, so maybe we don't have need to re-implement it ?

muneebahmedayub: @IAmTamal yeah it is default. But I am thinking about changing it to revalidation, what do you suggest how much time should it take to revalidate?

IAmTamal: If the data isn't dynamic - I don't think we might need to add `revalidation` to it. By dynamic, I mean any data that would get changed via CMS, or even by the user themselves. Here I see we are returning blog contents so I don't think the data will be changed too often right ?

muneebahmedayub: What if user creates another blog. Shouldn't it appear on the website dynamically?

muneebahmedayub:
@saadpasta  Now the api calls are moved in `client` folder and helper functions are moved to `utils`

saadpasta (approved):
LGTM

src/containers/Blogs/Blogs.jsx
saadpasta: Let's not think about that much right now let's push this and we can a have a look at this later

saadpasta:
@allcontributors please add @muneebahmedayub for code

PR #736, opened 2024-10-11
Solved many of the open issues

saadrasheeddev:
In original repository,

- In small screens, navbar should close once we clicked on any element. but it stays open on top of page and takes half of the top screen.
- While we change theme in small screen, navbar should get closed.
- Buttons of header in small screens are weird

All these issues are solved in it
But since I was aiming to make it my portfolio, I made data changes as well. Keep this in mind if this gets merged


saadpasta:
@saadrasheeddev you have updated lot of thing that should not be updated including the other thing please open issue and tag issues that are fixed thanks

PR #766, opened 2025-01-23
Temp mobile improvement

rfig81:
Change name to Reynold Figueroa so that the page header doesn't break on mobile. Design adjustments will be made after.


rfig81:
This was a miskate, my  bad.

PR #824, opened 2025-10-15
Enhancement: Use external Google Drive link for resume

brennogf:
# Description

This enhancement replaces the local resume file with an external Google Drive link.  
The resume content can now be updated externally without modifying the source code.  
No additional dependencies are required.

## Type of change

- [x] New feature (non-breaking change which adds functionality)


brennogf:
The build was failing because Vercel was using Node 18, which is discontinued. I added "engines": { "node": "22.x" } to the package.json, and now it works locally and in CI with Node 22 (tested with Docker).

This should make the build pass once merged.

